PR #4, opened 2020-04-15
6101 add docker files

SimonLLawrence:


docker-compose-build.yml
```diff
@@ -0,0 +1,10 @@
+version: '3.5'
+
+services:
+
+  nhsd.buyingcatalogue.ordering.api:
+    image: ${REGISTRY:-nhsd}/buying-catalogue/ordering-isapi:${TAG:-latest}
```
assoomro, line 6: would you please consider changing the image name from:
 `${REGISTRY:-nhsd}/buying-catalogue/ordering-isapi:${TAG:-latest}`
to 
`${REGISTRY:-nhsd}/buying-catalogue/ordering-api:${TAG:-latest}`

Isapi is an abbreviation we've been using for 'Identity Server API' and might cause confusion.

docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+version: '3.5'
+services:
+  nhsd.buyingcatalogue.ordering.api:
+    image: ${REGISTRY:-nhsd}/buying-catalogue/ordering-isapi:${TAG:-latest}
```
assoomro, line 4: same comment as above ⬆️

src/NHSD.BuyingCatalogue.Ordering.Api/Dockerfile
```diff
@@ -0,0 +1,22 @@
+#See https://aka.ms/containerfastmode to understand how Visual Studio uses this Dockerfile to build your images for faster debugging.
+
```
assoomro, line 2: Would it be possible to remove these 2 lines? 

In my opinion they don't add any value to the Dockerfile itself.

PR #7, opened 2020-04-20
5482: API Identity Resource on Ordering

RuthBerry96:


src/NHSD.BuyingCatalogue.Ordering.Api/NHSD.BuyingCatalogue.Ordering.Api.csproj
```diff
@@ -20,6 +20,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="3.1.3" />
```
sam-roberts-bjss, line 23: Just need to be aware this is a newer version of the package compared with what's defined in the Organisations.Api project. No need to change it. If anything we should be updating all the packages within other projects. I think this will be a task for next sprint.

src/NHSD.BuyingCatalogue.Ordering.Api/Startup.cs
```diff
@@ -11,6 +12,8 @@ namespace NHSD.BuyingCatalogue.Ordering.Api
 {
     public sealed class Startup
     {
+        private const string BearerToken = "Bearer";
```
sam-roberts-bjss, line 15: I would have thought the JWT package would have a constant for this. Would you mind having a quick look to see if there is? No worries if not.

src/NHSD.BuyingCatalogue.Ordering.Api/NHSD.BuyingCatalogue.Ordering.Api.csproj
RuthBerry96: 👍

src/NHSD.BuyingCatalogue.Ordering.Api/Startup.cs
RuthBerry96: Of course. Will let you know 😄

RuthBerry96: Just pushed the change. JwtBearerDefaults.AuthenticationScheme is "Bearer"

PR #9, opened 2020-04-21
6101: Added initial database project for Ordering Database

SimonLLawrence:


docker-compose.yml
```diff
@@ -1,5 +1,30 @@
...
+      - MSSQL_PID=Express
+      - DB_NAME=${NHSD_LOCAL_DB_NAME:-CatalogueOrdering}
+      - NHSD_PASSWORD=DisruptTheMarket1!
+      - INSERT_TEST_DATA=True
+      - CREATE_EA_USER=True
+      - EA_USER_FIRST_NAME=Agency
+      - EA_USER_LAST_NAME=User
+      - EA_USER_EMAIL=user@agency.com
+      - EA_USER_PASSWORD_HASH=AQAAAAEAACcQAAAAEEOFkNNrFpKmDC2TBv2CP/dzxfnjdXk97RoqRlunE/CGs2tmFcewKZj4M/fITiP2tg==
+      - EA_USER_PHONE=01234567890
```
assoomro, line 24: I'm not quite sure that this will be relevant to the ordering db, I think it would be a good idea to get rid of these variables until we actually need them

src/NHSD.BuyingCatalogue.Ordering.OrderingDatabase/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+WORKDIR /usr/config
+
+COPY /ShellScripts .
+COPY /Deployment sql/Deployment
+COPY /Indexes sql/Indexes
+COPY /Tables sql/Tables
+COPY /Security sql/Security
+
+RUN chmod +x /usr/config/entrypoint.sh
+RUN chmod +x /usr/config/configure-db.sh
```
assoomro, line 13: This isn't anything major, but we could improve this by combining them into 1 run statement (to save docker the trouble of making a whole new layer just for a 'chmod'). Entirely up to you to make any changes based on my suggestion. [Some reading on this topic.](https://stackoverflow.com/questions/39223249/multiple-run-vs-single-chained-run-in-dockerfile-what-is-better)
 
So instead of 
```
RUN chmod +x /usr/config/entrypoint.sh
RUN chmod +x /usr/config/configure-db.sh
```
Combine them
```
RUN chmod +x /usr/config/entrypoint.sh && chmod +x /usr/config/configure-db.sh
```

SimonLLawrence:
Fixed the docker file  so it should now build and also corrected CRLF on entrypoint.sh

PR #10, opened 2020-04-22
6101 create build pipeline

assoomro:


azure-pipelines.yml
```diff
@@ -0,0 +1,159 @@
...
+      includePreviewVersions: false
+
+  - task: DockerInstaller@0
+    inputs:
+      dockerVersion: $(dockerVersion)
+
+  - script: docker-compose up --build -d
+    displayName: 'Docker up'
+
+  - task: DotNetCoreCLI@2
```
sam-roberts-bjss, line 69: Just wanted to point your attention at the following resource [Docker with integration and unit tests](https://medium.com/@JoeHonour/a-guide-to-setting-up-a-net-core-project-using-docker-with-integrated-unit-and-component-tests-a326ca5a0284) as I think it maybe useful to run the unit tests from within the container. I have no evidence to base this off but I assume by running the run tests from the project would apply another restore & a build that should have already been carried out within the docker build. I think an improvement could be made here & potentially speed up the build process.

assoomro: Thank you for bringing this up. 

In essence, I think you're right with those assumptions and you're also right that it would speed up the build process.

However, this would hinder the development process, as running all tests when spinning up your environment would slow down the feedback loop, not to mention possible cases when you won't even be able to build due to failing tests.

So we have 2 things to weigh against each other:

| PROS                                      | CONS                       |
| :------------------------------: | :---------------------: |
| Faster completion of the pipeline process | Slower development process |

My opinion is that the overall time saved by faster build process will *not* outweigh the time lost on running tests locally when spinning up the environment.

I think it might be a good idea to get opinions from more people and see where the consensus lies and take action based on that.

What do you think? @sam-roberts-bjss

PR #11, opened 2020-04-23
6101: added template README.md 

SimonLLawrence:
Initial read me file to be updated as the service is developed

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,12 +17,15 @@ This application uses **.NET core** to provide an ordering API.
 
 ### Project structure
 
-This repository uses **.NET Core**, **Nodejs** and **Docker**.
+This repository uses **.NET Core**,  **Docker**.
 
 It contains the following endpoints:
 
-- ordering/order
-  - Returns a HTML view.
+- health/live
+  - Returns the health status of the service.
+
+- health/ready
+  - Returns the health status of the service dependacies  
 
 The application is broken down into the following project libraries:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,2 +1,78 @@
...
+To start up the web application, run the following command from the root directory of the repository.
+
+```shell
+docker-compose up -d --build
+```
+
+This will start the application in a docker container. You can verify that the service has launched correctly by navigating to the following url via any web browser.
+
+```http
+http://localhost:XXX/ordering/Order
```
sam-roberts-bjss, line 65: Regarding the URLs of the application I think the only endpoints we have at the moment are the health check ones. What do you think to changing the locations to point at those instead?

E.g.
You can verify that the API has launched correctly by navigating to the following URLs via any web browser.

http://localhost:xxxx//health/live
http://localhost:xxxx//health/ready

If both URLs return 'Healthy', the environment is configured correctly, and can be accessed via the public endpoints.

If the ready URL returns 'Unhealthy', the associated dependencies of the application may have failed to launch, or cannot be accessed.

README.md
```diff
@@ -1,2 +1,78 @@
...
+
+
+##### How to connect
+
+| From                       | Host                       | Port  | TLS |
+|            :-:             |            :-:             |  :-:  | :-: |
+| outside the docker network | localhost                  | 5104  |  X  |
+
+
+Navigate yourself to [localhost:5104](http://localhost:5104/) to view the Ordering API
```
sam-roberts-bjss, line 52: Please see comment below regarding the URLs of the application.

PR #12, opened 2020-05-06
4826: Canned Data - GET

RuthBerry96:


src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrdersController.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using NHSD.BuyingCatalogue.Ordering.Api.Models;
+
+namespace NHSD.BuyingCatalogue.Ordering.Api.Controllers
+{
+    [Route("api/v1/[controller]")]
+    [ApiController]
+    [Produces("application/json")]
+    [AllowAnonymous]
```
sam-roberts-bjss, line 12: Obviously we need to ensure we don't allow Anonymous users in the next implementation, but this is fine for now. We do already have the policies in place, so should be easy to add either in this PR or the next.

RuthBerry96: Yep. I just had allowAnoymous for the canned data 😄

PR #16, opened 2020-05-07
Change config to work with cluster by default

assoomro:
[AB#6827](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/6827)


src/NHSD.BuyingCatalogue.Ordering.Api/Properties/launchSettings.json
```diff
@@ -9,12 +9,14 @@
...
-      "commandName": "Docker",
+    "NHSD.BuyingCatalogue.Ordering.Api Docker": {
+      "commandName": "Project",
       "launchBrowser": true,
-      "launchUrl": "{Scheme}://{ServiceHost}:{ServicePort}/buyingcatalogueordering",
-      "publishAllPorts": true,
-      "useSSL": true
+      "environmentVariables": {
+        "ASPNETCORE_ENVIRONMENT": "Development",
+        "ConnectionStrings__OrderingDb": "Data Source=localhost,1533;Initial Catalog=CatalogueOrdering;MultipleActiveResultSets=true;User ID=NHSD;Password=DisruptTheMarket1!"
```
sam-roberts-bjss, line 17: Do you think it would be a good idea to use a different port for this database so it wouldn't clash with other parts of product suite e.g. Identity?

src/NHSD.BuyingCatalogue.Ordering.Api/Properties/launchSettings.json
```diff
@@ -9,12 +9,14 @@
...
+      },
+      "applicationUrl": "http://localhost:5104"
     }
   }
 }
```
sam-roberts-bjss, line 22: New line?

src/NHSD.BuyingCatalogue.Ordering.Api/appsettings.json
```diff
@@ -7,7 +7,7 @@
     }
   },
   "ConnectionStrings": {
-    "OrderingDb": "Data Source=localhost,1533;Initial Catalog=CatalogueOrdering;MultipleActiveResultSets=true;User ID=NHSD;Password=DisruptTheMarket1!"
+    "OrderingDb": "Data Source=localhost,1452;Initial Catalog=CatalogueOrdering;MultipleActiveResultSets=true;User ID=NHSD;Password=DisruptTheMarket1!"
```
sam-roberts-bjss, line 10: Please ensure the name of the database is consistent with the cluster database name as when attempting to use the local cluster the database name we set to 'buyingcatalogue'.

assoomro: Good shout! I've changed the name as part of [this PR](https://github.com/nhs-digital-gp-it-futures/platform/pull/119). I'm planning merge them at the same time to not induce issues with the local cluster setup :)

sam-roberts-bjss: 👍

src/NHSD.BuyingCatalogue.Ordering.Api/Properties/launchSettings.json
assoomro: 👍

assoomro: 👍

src/NHSD.BuyingCatalogue.Ordering.Api/Properties/launchSettings.json
assoomro: We could change it to a different port, however, that will take away the convenience when trying to log in to your [SSMS](https://docs.microsoft.com/en-us/sql/ssms/sql-server-management-studio-ssms?view=sql-server-ver15) (or its alternatives, if there are any) to debug your db when running integration tests.

When looking at it - we'd be giving away some comfort to prevent an edge case (this is assuming that nobody is standing up multiple integration test environments at the same time).

Given that my assumption is valid, I'd lean towards keeping it on port 1533. What do you think?

sam-roberts-bjss: I don't see it as an inconvenience to change the port of the database as it is probably an edge case having to connect using SSMS locally to the DB as well, once it is setup, should just work. At least with a different port I can ensure the two environments are separate & I can be pretty confident some other database is not causing any false positives in the Ordering.Api tests. This would also be consistent with the other applications that ensure all ports are unique for their environment.

assoomro: I see your point. I'll change it.

PR #20, opened 2020-05-12
5322: GET & PUT canned data - description

RuthBerry96:
[AB#6889](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/6889)

### DIFF
--- a/src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrderDescriptionController.cs
+++ b/src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrderDescriptionController.cs
@@ -1,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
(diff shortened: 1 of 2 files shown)


src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrderDescriptionController.cs
```diff
@@ -0,0 +1,40 @@
...
+        public ActionResult Update([FromRoute][Required]string orderId, [FromBody][Required] OrderDescriptionModel model)
+        {
+            if (model is null)
+                throw new ArgumentNullException(nameof(model));
+
+            _cannedData[orderId] = model.Description;
+            return NoContent();
+        }
+
+        private static readonly Dictionary<string, string> _cannedData = new Dictionary<string, string>();
```
sam-roberts-bjss, line 38: I believe after our coding standards session we agreed to place member variables at the top of the class.

RuthBerry96: Its being changed as I've started the GET implementation, so not much point in removing it. Noted though for code written at later dates

PR #30, opened 2020-05-18
6766: Added Authentication of Organisation ID to user claims

CallumBradburyBJSS:


src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrdersController.cs
```diff
@@ -29,6 +30,13 @@ public OrdersController(IOrderRepository orderRepository)
         [Route("/api/v1/organisations/{organisationId}/[controller]")]
         public async Task<ActionResult> GetAllAsync(Guid organisationId)
         {
+            var primaryOrganisationId = User.GetPrimaryOrganisationId();
```
sam-roberts-bjss, line 33: Might want to change this to return a GUID as this is more inline with the type of the user's ID?

src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrdersController.cs
```diff
@@ -29,6 +30,13 @@ public OrdersController(IOrderRepository orderRepository)
...
+
+            if (primaryOrganisationId != organisationId.ToString())
```
sam-roberts-bjss, line 35: By changing the above type to a GUID I believe this would mean don't have to worry about case?

src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrderDescriptionController.cs
```diff
@@ -28,11 +29,16 @@ public OrderDescriptionController(IOrderRepository orderRepository)
         public async Task<ActionResult> GetAsync([FromRoute][Required] string orderId)
         {
             var order = await _orderRepository.GetOrderByIdAsync(orderId);
+            var primaryOrganisationId = User.GetPrimaryOrganisationId();
```
sam-roberts-bjss, line 32: Only need to get the primary organisation ID if an order is returned. Could this be moved closer to its usage i,e, Just before the line `if (primaryOrganisationId != order.OrganisationId.ToString())`? I think this helps readability keeping context close together.

src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrderDescriptionController.cs
```diff
@@ -51,11 +57,16 @@ public async Task<ActionResult> UpdateAsync([FromRoute][Required]string orderId,
             }
 
             var order = await _orderRepository.GetOrderByIdAsync(orderId);
+            var primaryOrganisationId = User.GetPrimaryOrganisationId();
```
sam-roberts-bjss, line 60: Same applies here. Create the variable primaryOrganisationId closer to usage?

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Order Update Description 5322.feature
```diff
@@ -62,7 +62,14 @@ Scenario: 6. A non buyer user cannot update an orders description
         | Another Description |
 
 @5322
-Scenario: 7. Service Failure
+Scenario: 7. A buyer user cannot update an orders description for an organisation they don't belong to
+     Given the user is logged in with the Buyer role for organisation e6ea864e-ef1b-41aa-a4d5-04fc6fce0933
+    When the user makes a request to update the description with the ID C000014-01
+        | Description         |
+        | Another Description |
```
sam-roberts-bjss, line 69: Do we need an additional check here to validate the status code of the response, ensuring it is forbidden?

src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrdersController.cs
CallumBradburyBJSS: oh nice, you're right

PR #33, opened 2020-05-19
5322: PUT - name of the user

RuthBerry96:
Story: [AB#5322](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/5322)
Task: [AB#7050](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7050)

### DIFF
--- a/src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrderDescriptionController.cs
+++ b/src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrderDescriptionController.cs
@@ -2,7 +2,6 @@
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using System.Threading.Tasks;
-using IdentityModel;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using NHSD.BuyingCatalogue.Ordering.Api.Extensions;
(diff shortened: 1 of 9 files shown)


src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrderDescriptionController.cs
```diff
@@ -75,8 +76,12 @@ public async Task<ActionResult> UpdateAsync([FromRoute][Required]string orderId,
                 return BadRequest(new ErrorsModel(isValid.Errors.Select(x => new ErrorModel(x.Id, x.Field))));
             }
 
+            var name = User.FindFirst(x => x.Type == JwtClaimTypes.Name).Value;
```
sam-roberts-bjss, line 79: Rather than bringing in the identity package just for the claim type name. Apparently there is the property `User.Identity.Name` to retrieve the name of the user. However by default Identity looks for a particular namespace. I found [this](https://www.jerriepelser.com/blog/overriding-name-claimtype-aspnetcore-oidc/) resource that explains it. What do you think?

@SimonLLawrence Making sure you are aware of this as well.

RuthBerry96: 👍

PR #39, opened 2020-05-20
Feature/4616 put canned data

RuthBerry96:
Story: [AB#4616](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/4616)
Task: [AB#7169](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7169)


src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrderingPartyController.cs
```diff
@@ -0,0 +1,63 @@
...
+                    EmailAddress = "bob.smith@email.com",
+                    TelephoneNumber = "4342345223  3434324"
+                }
+            };
+
+            return Ok(result);
+        }
+
+        [HttpPut]
+        public ActionResult Update([FromRoute][Required]string orderId, [FromBody][Required] OrderingPartyModel model)
```
bjss-chris, line 48: None of the parameter attributes are needed and can be safely removed.

src/NHSD.BuyingCatalogue.Ordering.Api/Models/AddressModel.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace NHSD.BuyingCatalogue.Ordering.Api.Models
+{
+    public sealed class AddressModel
```
bjss-chris, line 3: Class can be `internal`.

src/NHSD.BuyingCatalogue.Ordering.Api/Models/OrderingPartyModel.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace NHSD.BuyingCatalogue.Ordering.Api.Models
+{
+    public sealed class OrderingPartyModel
```
bjss-chris, line 3: Class can be `internal`.

src/NHSD.BuyingCatalogue.Ordering.Api/Models/OrganisationModel.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace NHSD.BuyingCatalogue.Ordering.Api.Models
+{
+    public sealed class OrganisationModel
```
bjss-chris, line 3: Class can be `internal`.

src/NHSD.BuyingCatalogue.Ordering.Api/Models/PrimaryContactModel.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace NHSD.BuyingCatalogue.Ordering.Api.Models
+{
+    public sealed class PrimaryContactModel
```
bjss-chris, line 3: Class can be `internal`.

src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrderingPartyController.cs
RuthBerry96: 👍

src/NHSD.BuyingCatalogue.Ordering.Api/Models/OrderingPartyModel.cs
RuthBerry96: 👍

src/NHSD.BuyingCatalogue.Ordering.Api/Models/OrganisationModel.cs
RuthBerry96: 👍

src/NHSD.BuyingCatalogue.Ordering.Api/Models/PrimaryContactModel.cs
RuthBerry96: 👍

src/NHSD.BuyingCatalogue.Ordering.Api/Models/AddressModel.cs
RuthBerry96: They need to be public, or in the contructor in the update method it errors. I can either make two different models (update and get) with the get internal and update public, but seen as they have exactly the same content it seems a bit pointless

bjss-chris: Yes, makes sense – will have been fine as `internal` until you introduced the put.

PR #47, opened 2020-05-21
7205: Added supplier data columns to order table

SimonLLawrence:
[AB#7205](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7205)  added supplier data columns to order table

### DIFF
--- a/src/NHSD.BuyingCatalogue.Ordering.OrderingDatabase/Tables/Contact.sql
+++ b/src/NHSD.BuyingCatalogue.Ordering.OrderingDatabase/Tables/Contact.sql
@@ -1,6 +1,6 @@
 ﻿CREATE TABLE dbo.Contact
 (
-    [ContactId] INT IDENTITY(1, 1) NOT NULL,
+    ContactId INT IDENTITY(1, 1) NOT NULL,
     FirstName NVARCHAR(100) NULL, 
     LastName NVARCHAR(100) NULL,
     Email NVARCHAR(100) NULL, 
--- a/src/NHSD.BuyingCatalogue.Ordering.OrderingDatabase/Tables/Order.sql
+++ b/src/NHSD.BuyingCatalogue.Ordering.OrderingDatabase/Tables/Order.sql
@@ -1,13 +1,18 @@
 ﻿CREATE TABLE dbo.[Order]
 (
     OrderId NVARCHAR(10) NOT NULL PRIMARY KEY,
-    [Description] NVARCHAR(100) NOT NULL,
+    Description NVARCHAR(100) NOT NULL,
     OrganisationId UNIQUEIDENTIFIER NOT NULL,
     OrderStatusId INT NOT NULL,
+    SupplierId INT NULL, 
+    SupplierName VARCHAR(255) NULL, 
+    SupplierAddressId INT NULL, 
+    SupplierContactId INT NULL,
     Created DATETIME2 NOT NULL,
     LastUpdated DATETIME2 NOT NULL,
     LastUpdatedBy UNIQUEIDENTIFIER NOT NULL,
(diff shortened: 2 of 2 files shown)


src/NHSD.BuyingCatalogue.Ordering.OrderingDatabase/Tables/Contact.sql
```diff
@@ -1,9 +1,9 @@
 ﻿CREATE TABLE dbo.Contact
 (
-    [ContactId] INT IDENTITY(1, 1) NOT NULL,
-    FirstName VARCHAR(100) NULL, 
+    ContactId INT IDENTITY(1, 1) NOT NULL,
```
RuthBerry96, line 3: Thanks for changing them, thought I did them all (nvarchar) :)

src/NHSD.BuyingCatalogue.Ordering.OrderingDatabase/Tables/Order.sql
```diff
@@ -1,20 +1,25 @@
...
-    [Description] NVARCHAR(100) NOT NULL,
+    Description NVARCHAR(100) NOT NULL,
     OrganisationId UNIQUEIDENTIFIER NOT NULL,    
     OrganisationName NVARCHAR(256) NULL, 
     OrganisationOdsCode NVARCHAR(8) NULL, 
     OrganisationAddressId INT NULL,
     OrganisationBillingAddressId INT NULL,    
     OrganisationContactId INT NULL,
     OrderStatusId INT NOT NULL,    
+    SupplierId INT NULL,
```
RuthBerry96, line 12: Really good, would you be able to put this in the Order table (if you have a look at the latest PR, I've setup foreign keys using the data annotations, as otherwise this broke the integration tests as EF was struggling to map

SimonLLawrence: Not sure what you are saying here do you want me to add the attributes to the data model?

RuthBerry96: To the Order.cs (in the domain)

RuthBerry96: This is an example of linking the foreign key to it

RuthBerry96: [ForeignKey("OrganisationAddressId")]
        public Address OrganisationAddress { get; set; }

From https://github.com/nhs-digital-gp-it-futures/BuyingCatalogueOrdering/pull/48/files

SimonLLawrence: ok your comment said order table , but you i think you mean i should add he foreign key attributes to the orders model? ,  also SupplierId does not map to a table row at the moment so i guess you mean just  the attributes for the other relations  address and contact?

RuthBerry96: Yes sorry, use the foreign key attribute on address and contact, but also include the supplier name and supplierId on the order.cs class

SimonLLawrence: I need to patch tests if i do model stuff now, i kind of think the model stuff should have been done in a separate commit .  now the column commit will have model changes and test changes too.

SimonLLawrence: Going to do the model changes in a separate commit

PR #56, opened 2020-05-27
4616: Empty result bug

RuthBerry96:
Story: [AB#4616](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/4616)
Task: [AB#7371](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7371)

### DIFF
--- a/src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrderingPartyController.cs
+++ b/src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrderingPartyController.cs
@@ -44,7 +44,7 @@ namespace NHSD.BuyingCatalogue.Ordering.Api.Controllers
 
             var result = new OrderingPartyModel
             {
-                Organisation = primaryContact is null ? null : new OrganisationModel
+                Organisation = !order.IsOrderingPartySectionComplete() ? null : new OrganisationModel
                 {
                     Name = order.OrganisationName,
                     OdsCode = order.OrganisationOdsCode,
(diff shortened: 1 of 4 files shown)


src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrderingPartyController.cs
```diff
@@ -44,7 +44,7 @@ public async Task<ActionResult<OrderingPartyModel>> GetAsync(string orderId)
 
             var result = new OrderingPartyModel
             {
-                Organisation = new OrganisationModel
+                Organisation = primaryContact is null ? null : new OrganisationModel
```
bjss-chris, line 47: Do we not want to return anything if we have no primary contact details or is having no primary contact details not considered a valid scenario?

sam-roberts-bjss: I think the only mandatory data we can be sure of is the primary contact information as this is entered through the order form application. The Organisation information is retrieved via the ODS service. Can we be sure the address information will be set? The basic requirement at the moment is if the Front end would like an empty object if no ordering party information is set. This is used to determine whether or not to make a request to the Organisation API.

bjss-chris: As discussed on the call, I think it might be worth refactoring this check to a method with a name like `HasMandatoryData` or `IsSectionComplete` to hopefully provide some additional context as to why we return nothing if we have no primary contact details.

sam-roberts-bjss: 👍  Like the giving additional context approach by including a method name.

RuthBerry96: 👍

PR #70, opened 2020-06-01
Feature/4619 update order summary

sam-roberts-bjss:
Story: [AB#4619](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/4619)
Task: [AB#7408](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7408)


tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Builders/OrderBuilder.cs
```diff
@@ -15,27 +15,15 @@ private OrderBuilder()
...
-                    Postcode = "LS1 1SX"
-                },
-                SupplierContact = new Contact
-                {
-                    FirstName = "Sam",
-                    LastName = "Smith",
-                    Email = "sam.smith@email.com",
-                    Phone = "0123456789"
-                }
+                SupplierAddress = AddressBuilder.Create().WithLine1("1 Some Supplier").Build()
```
RuthBerry96, line 26: Do you not want to keep the supplierContact in? Is there a reason you have removed it and not put anything else in its place?

RuthBerry96: After our phone call it makes sense 😄

PR #81, opened 2020-06-03
7412: Add service recipients table (& amend canned data)

RuthBerry96:
Story: [AB#7412](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7412)
Task: [AB#7489](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7489)

### DIFF
--- a/src/NHSD.BuyingCatalogue.Ordering.OrderingDatabase/Tables/ServiceRecipient.sql
+++ b/src/NHSD.BuyingCatalogue.Ordering.OrderingDatabase/Tables/ServiceRecipient.sql
@@ -4,6 +4,6 @@
     OrderId NVARCHAR(10) NOT NULL,
     [Name] NVARCHAR(256) NULL, 
     OdsCode NVARCHAR(8) NULL, 
-    CONSTRAINT PK_ServiceRecipient PRIMARY KEY (ServiceRecipientId),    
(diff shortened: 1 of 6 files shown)


src/NHSD.BuyingCatalogue.Ordering.OrderingDatabase/Tables/ServiceRecipient.sql
```diff
@@ -0,0 +1,9 @@
+﻿CREATE TABLE dbo.ServiceRecipient
+(
+    ServiceRecipientId INT IDENTITY(1, 1) NOT NULL,
+    OrderId NVARCHAR(10) NOT NULL,
+    [Name] NVARCHAR(256) NULL, 
+    OdsCode NVARCHAR(8) NULL, 
+    CONSTRAINT PK_ServiceRecipient PRIMARY KEY (ServiceRecipientId),
+    CONSTRAINT FK_ServiceRecipient_Order FOREIGN KEY (OrderId) REFERENCES [Order] (OrderId)
```
sam-roberts-bjss, line 8: Normally I would include the column name in the constraint as well. But I don't think this was mentioned in the coding standards meeting, so welcome to ignore. Just thinking could there ever be a case when we have two constraints to the same table but different fields, probably unlikely

PR #88, opened 2020-06-04
Feature/7412 get service recipients

RuthBerry96:
Story: [AB#7412](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7412)
Task: [AB#7467](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7467)


src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/ServiceRecipientsSectionController.cs
```diff
@@ -1,57 +1,76 @@
...
+                OdsCode = recipient.OdsCode,
+                Name = recipient.Name
+            }).ToList();
+
+            var model = new ServiceRecipientsModel
+            {
+                ServiceRecipients = recipientModelList
+            };
+
+            return model;
```
sam-roberts-bjss, line 62: Any reason to not return an OK(model)? Just thinking for consistency reasons.

src/NHSD.BuyingCatalogue.Ordering.Application/Persistence/IServiceRecipientRepository.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+using NHSD.BuyingCatalogue.Ordering.Domain;
+
+namespace NHSD.BuyingCatalogue.Ordering.Application.Persistence
+{
+    public interface IServiceRecipientRepository
+    {
+        Task<IEnumerable<ServiceRecipient>> ListServiceRecipientsByOrderId(string orderId);
```
sam-roberts-bjss, line 9: Async?

src/NHSD.BuyingCatalogue.Ordering.Domain/ServiceRecipient.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.ComponentModel.DataAnnotations;
+using System.ComponentModel.DataAnnotations.Schema;
+
+namespace NHSD.BuyingCatalogue.Ordering.Domain
+{
+    public sealed class ServiceRecipient
+    {
+        [Key]
```
sam-roberts-bjss, line 8: Shouldn't need these attributes. Would you mind reviewing the use of the Entity Framework mapping configuration, should be an option here, which would free the domain object from persistence concerns?

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/ServiceRecipients GetAll.feature
```diff
@@ -0,0 +1,54 @@
+﻿Feature: Display all of the Service Recipients for an order
+    As a Buyer
+    I want to view all of the service recipients for an order
+    So that I can ensure the information is correct
+
+Background:
+    Given Orders exist
+        | OrderId    | Description         | OrderStatusId | LastUpdatedByName | LastUpdatedBy                        | OrganisationId                       |
```
sam-roberts-bjss, line 8: To reduce the amount of information & to hopefully make the code easier to read some of the properties have been defaulted e.g. `OrderStatusId`

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Builders/ServiceRecipientBuilder.cs
```diff
@@ -0,0 +1,34 @@
...
+    internal sealed class ServiceRecipientBuilder
+    {
+        private readonly ServiceRecipient _serviceRecipient;
+
+        private ServiceRecipientBuilder()
+        {
+            _serviceRecipient = new ServiceRecipient
+            {
+                Name = "Some name",
+                Order = new Order()
```
sam-roberts-bjss, line 14: Would like to suggest the use of the OrderBuilder class here rather than creating the Order class directly. If the Order constructor changed we just want that to impact one area of the tests, not multiple.

Also the other builder pattern might be a better fit here. Specify each property of the `ServiceRecipient` separately and create the object within the `Build` method. Might be best speaking about this one & seeing what you think but use the `AddressBuilder` as an example.

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Controllers/ServiceRecipientsSectionControllerTests.cs
```diff
@@ -0,0 +1,181 @@
...
+    [TestFixture]
+    [Parallelizable(ParallelScope.All)]
+    internal sealed class ServiceRecipientsSectionControllerTests
+    {
+        [Test]
+        public void Ctor_NullRepository_Throws()
+        {
+            Assert.Throws<ArgumentNullException>(() =>
+            {
+                var _ = new ServiceRecipientsSectionController(null, null);
```
sam-roberts-bjss, line 28: Not sure which constructor parameter this test is trying to verify. It might be clearer to have two tests one that supplies a null for the first parameter & the other test that supplies a null to the second. This way if a test fails it should be obvious which parameter didn't throw the correct exception.

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Controllers/ServiceRecipientsSectionControllerTests.cs
```diff
@@ -0,0 +1,181 @@
...
+            const string orderId = "C0000014-01";
+
+            var serviceRecipients = new List<(ServiceRecipient serviceRecipient, ServiceRecipientModel expectedModel)>
+            {
+                CreateServiceRecipientData("ODS1", orderId)
+            };
+
+            context.ServiceRecipients = serviceRecipients.Select(x => x.serviceRecipient).ToList();
+
+            var expected = new ServiceRecipientsModel();
```
sam-roberts-bjss, line 80: Possible to use an object initializer here?

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Controllers/ServiceRecipientsSectionControllerTests.cs
```diff
@@ -0,0 +1,181 @@
...
+                OrderRepositoryMock.Setup(x => x.GetOrderByIdAsync(It.IsAny<string>())).ReturnsAsync(() => Order);
+
+                ServiceRecipientRepositoryMock = new Mock<IServiceRecipientRepository>();
+                ServiceRecipients = new List<ServiceRecipient>();
+                ServiceRecipientRepositoryMock.Setup(x => x.ListServiceRecipientsByOrderId(It.IsAny<string>()))
+                    .ReturnsAsync(() => ServiceRecipients);
+
+                ClaimsPrincipal = new ClaimsPrincipal(new ClaimsIdentity(new[]
+                {
+                    new Claim("Ordering", "Manage"),
```
sam-roberts-bjss, line 145: @CallumBradburyBJSS Did you end up creating a builder for this common section of code?

sam-roberts-bjss:
My own fault but would you mind looking at some of the changes that looks like the whole file has changed & make sure the line endings are set to `LF` please? e.g. `ServiceRecipientsSectionController`

src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/ServiceRecipientsSectionController.cs
RuthBerry96: This was in response to one of @bjss-chris previous PR reviews, stating that you dont actually need to return Ok(model), but by returning model it still returns a 200. I can change it back if you prefer, but it is also in other areas of the code

src/NHSD.BuyingCatalogue.Ordering.Application/Persistence/IServiceRecipientRepository.cs
RuthBerry96: 👍

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/ServiceRecipients GetAll.feature
RuthBerry96: 👍

src/NHSD.BuyingCatalogue.Ordering.Domain/ServiceRecipient.cs
RuthBerry96: Added a key mapping to the application db context, however I had a problem with the foreign key attribute working

src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/ServiceRecipientsSectionController.cs
bjss-chris: @sam-roberts-bjss there's no benefit to returning `Ok(model)` – it just adds complexity to the code and tests that isn't needed and the result is exactly the same. However, it relies on people making sure they are using the `ActionResult<T>` return type rather than `ActionResult` or `IActionResult` but we should be doing this anyway for the reasons outlined [here](https://docs.microsoft.com/en-us/aspnet/core/web-api/action-return-types?view=aspnetcore-3.1#actionresultt-type) by MS.

We can always discuss this afternoon but my recommendation would be for @RuthBerry96 to leave this as is.

src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/ServiceRecipientsSectionController.cs
```diff
@@ -1,57 +1,76 @@
...
+            var order = await _orderRepository.GetOrderByIdAsync(orderId);
+
+            if (order is null)
+            {
+                return NotFound();
+            }
+
+            var serviceRecipients = (await _serviceRecipientRepository.ListServiceRecipientsByOrderId(orderId)).ToList();
+
+            var primaryOrganisationId = User.GetPrimaryOrganisationId();
```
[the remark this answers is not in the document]
RuthBerry96: 👍

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Controllers/ServiceRecipientsSectionControllerTests.cs
CallumBradburyBJSS: I didn't, I was going to bring it up at the code meeting this afternoon and get thoughts

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Controllers/ServiceRecipientsSectionControllerTests.cs
RuthBerry96: 👍

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Controllers/ServiceRecipientsSectionControllerTests.cs
RuthBerry96: 👍

src/NHSD.BuyingCatalogue.Ordering.Persistence/Data/ApplicationDbContext.cs
```diff
@@ -27,6 +27,7 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             base.OnModelCreating(modelBuilder);
 
             modelBuilder.ApplyConfiguration(new OrderConfiguration());
+            modelBuilder.Entity<ServiceRecipient>(x => x.HasKey(y => y.OdsCode));
```
sam-roberts-bjss, line 30: Could the ServiceRecipient configuration be encapsulated into an entity configuration class to be consistent with the `OrderConfiguration` class above?

RuthBerry96: 👍

PR #90, opened 2020-06-08
Feature/4629 add service recipient count

sam-roberts-bjss:
Story: [AB#4629](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/4629)
Task: [AB#7473](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7473)


tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Controllers/OrdersControllerTests.cs
```diff
@@ -25,12 +25,51 @@ namespace NHSD.BuyingCatalogue.Ordering.Api.UnitTests.Controllers
     internal sealed class OrdersControllerTests
     {
         [Test]
-        public void Constructor_NullRepository_Throws()
+        public void Constructor_Null_Service_Repository_ThrowsArgumentNullException()
```
RuthBerry96, line 28: Would it be possible to use a TestCase here, as these three are very similar. If you dont think it's worth it then dont worry

src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrdersController.cs
```diff
@@ -24,28 +24,29 @@ public sealed class OrdersController : Controller
     {
         private readonly IOrderRepository _orderRepository;
         private readonly ICreateOrderService _createOrderService;
+        private readonly IServiceRecipientRepository _serviceRecipientRepository;
```
RuthBerry96, line 27: I know you havent changed this in this PR, but is it worth changing the ":Controller" to ":ControllerBase", which most of the other controllers are now using. Then you can get rid of the "using" keyword in the tests

sam-roberts-bjss: 👍

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Controllers/OrdersControllerTests.cs
sam-roberts-bjss: Torn on this one. I've made the change but not sure I like it as find it harder to read which test may have caused the exception. Will try it out as it does simplify the test by reducing the amount of code.

PR #105, opened 2020-06-24
7982: Check db is running on integration tests

RuthBerry96:
Story: [AB#7827](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7827)
Task: [AB#7982](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7982)

### DIFF
--- a/tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Hooks/IntegrationHook.cs
+++ b/tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Hooks/IntegrationHook.cs
@@ -14,13 +14,13 @@ namespace NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests.Hooks
     public sealed class IntegrationHook
     {
         private readonly IObjectContainer _objectContainer;
-        private readonly DatabaseHealthCheck _databaseHealthCheck;
+        private readonly OrderingApiHealthCheck _orderingApiHealthCheck;
         private static bool _firstScenario = true;
 
-        public IntegrationHook(IObjectContainer objectContainer, DatabaseHealthCheck databaseHealthCheck)
+        public IntegrationHook(IObjectContainer objectContainer, OrderingApiHealthCheck orderingApiHealthCheck)
         {
             _objectContainer = objectContainer ?? throw new ArgumentNullException(nameof(objectContainer));
-            _databaseHealthCheck = databaseHealthCheck;
+            _orderingApiHealthCheck = orderingApiHealthCheck;
         }
 
         [BeforeScenario]
@@ -32,7 +32,7 @@ namespace NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests.Hooks
             if (_firstScenario)
             {
                 _firstScenario = false;
-                await _databaseHealthCheck.AwaitApiRunningAsync(_objectContainer.Resolve<Settings>());
+                await _orderingApiHealthCheck.AwaitApiRunningAsync(_objectContainer.Resolve<Settings>());
             }
 
             await ResetDatabaseAsync();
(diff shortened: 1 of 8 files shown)


tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Hooks/IntegrationHook.cs
```diff
@@ -25,6 +29,12 @@ public async Task BeforeScenarioAsync()
             RegisterTestConfiguration();
             RegisterCustomValueRetrievers();
 
+            if (_firstScenario)
```
sam-roberts-bjss, line 32: Shouldn't need this boolean if we use the attribute [BeforeTestRun]. Put this on top of a new method & it will be called before all tests run.

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Hooks/IntegrationHook.cs
```diff
@@ -13,10 +14,13 @@ namespace NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests.Hooks
     public sealed class IntegrationHook
     {
         private readonly IObjectContainer _objectContainer;
+        private readonly DatabaseHealthCheck _databaseHealthCheck;
```
sam-roberts-bjss, line 17: The name of the class `DatabaseHealthCheck` is a little misleading here as we are checking the health endpoints of the Ordering API which just so happens to check the database. If we added a new dependency to the Ordering Api this might add more confusion here. Off the top of my head for a better name, how about `OrderingApiHealthCheck`?

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Support/HttpClientAwaiter.cs
```diff
@@ -0,0 +1,36 @@
...
+        public static async Task<bool> WaitForGetAsync(string address, TimeSpan testTimeout)
+        {
+            using (var httpClient = new HttpClient() { Timeout = TimeSpan.FromSeconds(1) })
+            {
+                DateTime startTime = DateTime.UtcNow;
+                while (DateTime.UtcNow - startTime < testTimeout)
+                {
+                    try
+                    {
+                        using var response = await httpClient.GetAsync(new Uri(address)).ConfigureAwait(false);
```
sam-roberts-bjss, line 18: Don't need the `ConfigureAwait(false)` anymore, so could be removed.

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Utils/Settings.cs
```diff
@@ -10,6 +10,7 @@ public Settings(IConfiguration config)
             ConnectionString = config.GetConnectionString("OrderingDb");
             OrderingApiBaseUrl = config.GetValue<string>("OrderingApiBaseUrl");
             Authority = config.GetValue<string>("Authority");
+            TestTimeOutInSeconds = config.GetValue<int>("TestTimeOutInSeconds");
```
sam-roberts-bjss, line 13: Could the name of this be improved. At first glance I am not sure what time out it is for. How about 'OrderingApiHealthCheckTimeout'?

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Support/HttpClientAwaiter.cs
RuthBerry96: 👍

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Hooks/IntegrationHook.cs
RuthBerry96: [BeforeTestRun] needs the method to be static, so if we used this then we would have to hardcode the baseUrl and the time expiry

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Hooks/IntegrationHook.cs
RuthBerry96: 👍

PR #109, opened 2020-06-29
Feature/7840 update order item

sam-roberts-bjss:
Story: [AB#7840](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7840)
Task: [AB#7859](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7859)


src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/CatalogueSolutionsController.cs
```diff
@@ -155,27 +158,50 @@ public ActionResult<CreateOrderItemModel> GetOrderItem(string orderId, string or
         }
 
         [HttpPut]
-        [Route("{orderItemId}")]
+        [Route("{orderItemId:int}")]
```
RuthBerry96, line 161: Did you try the route without the :int? Did it still work. I just dont think you need it as you're specifiying the parameter underneith (but I could be wrong)

src/NHSD.BuyingCatalogue.Ordering.Api/Models/UpdateOrderItemModel.cs
```diff
@@ -5,8 +5,11 @@ namespace NHSD.BuyingCatalogue.Ordering.Api.Models
     public class UpdateOrderItemModel
     {
         public DateTime? DeliveryDate { get; set; }
+
```
RuthBerry96, line 8: I know it wasnt there before, but can you add sealed to this class

src/NHSD.BuyingCatalogue.Ordering.Api/Models/UpdateOrderItemResponseModel.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Collections.Generic;
+
+namespace NHSD.BuyingCatalogue.Ordering.Api.Models
+{
+    public sealed class UpdateOrderItemResponseModel
+    {
+        public IEnumerable<ErrorModel> Errors { get; set; }
```
RuthBerry96, line 7: Would it not be better to have a generic list of errors, as wont this be done for lots of different methods

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Builders/UpdateCatalogueSolutionOrderItemRequestPayloadBuilder.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System;
+using NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests.Requests;
+using NHSD.BuyingCatalouge.Ordering.Api.Testing.Data.Data;
+
+namespace NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests.Builders
+{
+    internal sealed class UpdateCatalogueSolutionOrderItemRequestPayloadBuilder
+    {
+        private DateTime? _deliveryDate;
+        private TimeUnit _estimationPeriod;
```
RuthBerry96, line 10: Time unit can be null right? Do you need the '?' a the end?

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Catalogue Solutions Update.feature
```diff
@@ -1,49 +1,87 @@
...
+    And Order items exist
+        | OrderId    | OdsCode | CatalogueItemName | CatalogueItemType |
+        | C000014-01 | ODS1    | Order Item 1      | Solution          |
+        | C000014-01 | ODS2    | Order Item 2      | AdditionalService |
     And the user is logged in with the Buyer role for organisation 4af62b99-638c-4247-875e-965239cd0c48
 
-@4631
-Scenario: 1. Update an orders catalogue solutions
-    When the user makes a request to update the order catalogue solutions section with the ID C000014-01
+@7859
```
RuthBerry96, line 20: I thought we were still using story numbers rather than task numbers? Could be wrong though

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Builders/OrderBuilder.cs
```diff
@@ -1,131 +1,172 @@
...
+ using System.Collections.Generic;
+ using NHSD.BuyingCatalogue.Ordering.Domain;
 
 namespace NHSD.BuyingCatalogue.Ordering.Api.UnitTests.Builders
 {
     internal sealed class OrderBuilder
     {
-        private readonly Order _order;
+        private string _orderId;
+        private string _orderDescription;
```
RuthBerry96, line 10: For the record, I'm not a massive fan of changing our builders to this new structure, as I think it reads better the old way. I was going to bring it up last Fri, but then we postponed it. Up to you though

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Builders/OrderBuilder.cs
```diff
@@ -1,131 +1,172 @@
...
+        private Guid _organisationId;
+        private readonly string _organisationName;
```
RuthBerry96, line 12: Also seen as there is a OrderBuilder and OrderItemBuilder in the domain layer why do we need one here?

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Controllers/CatalogueSolutionsControllerTests.cs
```diff
@@ -332,6 +332,217 @@ public async Task CreateOrderItemAsync_CreateOrderItemService_CalledOnce()
                     x.CreateAsync(It.IsNotNull<CreateOrderItemRequest>()), Times.Once);
         }
 
+        [Test]
+        public void UpdateOrderItemAsync_NullModel_ThrowsArgumentNullException()
+        {
+            static async Task TestAsync()
+            {
+                const string orderId = "C10000-01";
+                int orderItemId = 1;
```
RuthBerry96, line 341: can you prefix the int with const

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Services/UpdateOrderItemServiceTests.cs
```diff
@@ -0,0 +1,182 @@
...
+                IdentityServiceMock.Setup(identityService => identityService.GetUserIdentity()).Returns(() => UserId);
+                IdentityServiceMock.Setup(identityService => identityService.GetUserName()).Returns(() => UserName);
+
+                UpdateOrderItemService = UpdateOrderItemServiceBuilder
+                    .Create()
+                    .WithOrderRepository(OrderRepositoryMock.Object)
+                    .WithIdentityService(IdentityServiceMock.Object)
+                    .Build();
+            }
+
```
RuthBerry96, line 162: New line can be removed

tests/NHSD.BuyingCatalouge.Ordering.Api.Testing.Data/Entities/OrderItemEntity.cs
```diff
@@ -84,7 +84,8 @@ public sealed class OrderItemEntity : EntityBase
                 @Price,
                 @Created,
                 @LastUpdated
-            );";
+            );
+            SELECT SCOPE_IDENTITY();";
```
RuthBerry96, line 88: Is there a reason why you have added this to the tests, just cant see where it is used to get the ID

src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/CatalogueSolutionsController.cs
sam-roberts-bjss: According to this [website](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-3.1#route-constraint-reference), route constraints should not be used for input validation. I will remove this constraint.

src/NHSD.BuyingCatalogue.Ordering.Api/Models/UpdateOrderItemResponseModel.cs
sam-roberts-bjss: Do you mean introduce a abstract Response class?

RuthBerry96: I think so. I depends how many time we are going to make a class thats a list of errors (which I imagine will be a fair bit)

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Builders/UpdateCatalogueSolutionOrderItemRequestPayloadBuilder.cs
sam-roberts-bjss: It is a reference type so there is no need to add a '?' as it can be null. DateTime by itself cannot be null & would use a default value.

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Builders/OrderBuilder.cs
sam-roberts-bjss: This was reviewed during a coding standards meeting & I thought this was what we agreed. Creating the object in the constructor makes it difficult to test nullability. Take a Controller or a Service class for instance. If the object is created in the constructor you cannot pass null in as a parameter as that will throw an exception. This means a builder cannot be used in an argument null exception test, which defeats the purpose of the builder to encapsulate the creation of an object to one place so if the constructor changes we only have to change it in the builder & not in a set of test methods. Welcome to bring it up again if you can see advantages for the other approach.

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Builders/OrderBuilder.cs
sam-roberts-bjss: Both the domain and api layers use the domain classes for test data. It would make sense to move these builders into a common library.

src/NHSD.BuyingCatalogue.Ordering.Api/Models/UpdateOrderItemResponseModel.cs
sam-roberts-bjss: Yeah. I think we keep an eye on it then & see if there is an appropriate abstraction. This could be one to bring up in the next coding standards meeting. If agreed, I will leave this for now.

stuart-bass-bjss: why do you need a specific response class anyway, if it's IEnumerable<ErrorModel>?

sam-roberts-bjss: What do you mean?

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Steps/CreateCatalogueSolutionOrderItemSteps.cs
```diff
@@ -38,13 +33,10 @@ public void GivenTheUserCreatesARequestToAddANewCatalogueSolutionOrderItemToTheO
                 orderId);
         }
 
-        [Given(@"the user enters '(.*)' payload")]
+        [Given(@"the user enters the '(.*)' create catalogue solution order item request payload")]
```
CallumBradburyBJSS, line 36: Glad to see this change! Was looking at this in dev this morning and reached the same conclusion you seem to have, here

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Controllers/CatalogueSolutionsControllerTests.cs
```diff
@@ -332,6 +332,217 @@ public async Task CreateOrderItemAsync_CreateOrderItemService_CalledOnce()
                     x.CreateAsync(It.IsNotNull<CreateOrderItemRequest>()), Times.Once);
         }
 
+        [Test]
+        public void UpdateOrderItemAsync_NullModel_ThrowsArgumentNullException()
+        {
+            static async Task TestAsync()
```
CallumBradburyBJSS, line 338: I'm not a fan of these static methods, as if anything in the test setup was to throw the exception, the test would pass, not only if the function we're wanting to call did. Prefer to do everything in the test as normal, then do

Assert.ThrowsAsync<ArgumentNullException(async () => context.Controller.UpdateOrderItemAsync(orderId, orderItemId, null)

Plus it lets us avoid nested functions which is always a bonus. Won't request a change, one for a code standards meeting though I think

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Services/UpdateOrderItemServiceTests.cs
```diff
@@ -0,0 +1,181 @@
...
+            static void Test()
+            {
+                UpdateOrderItemServiceBuilder
+                    .Create()
+                    .WithOrderRepository(null)
+                    .WithIdentityService(Mock.Of<IIdentityService>())
+                    .Build();
+            }
+
+            Assert.Throws<ArgumentNullException>(Test);
```
CallumBradburyBJSS, line 32: arguably here you can skip the static function and do:

```
var builder = UpdateOrderItemServiceBuilder
                    .Create()
                    .WithOrderRepository(null)
                    .WithIdentityService(Mock.Of<IIdentityService>());

Assert.Throws<ArgumentNullException>(builder.Build);
```


Which I think is a lot cleaner and again will only assert on the construction itself, not the setup.

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Requests/UpdateCatalogueSolutionOrderItemRequest.cs
```diff
@@ -0,0 +1,119 @@
...
+                    "complete", () =>
+                    UpdateCatalogueSolutionOrderItemRequestPayloadBuilder
+                        .Create()
+                        .Build()
+                },
+                {
+                    "high-boundary", () =>
+                    UpdateCatalogueSolutionOrderItemRequestPayloadBuilder
+                        .Create()
+                        .WithDeliveryDate(DateTime.MaxValue)
```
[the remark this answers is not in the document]
sam-roberts-bjss: Cheers 👍

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Controllers/CatalogueSolutionsControllerTests.cs
sam-roberts-bjss: Yeah one for the next coding standards meeting then. Happy to go with the consensus.

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Services/UpdateOrderItemServiceTests.cs
sam-roberts-bjss: Added to coding standards task for review:

https://buyingcatalog.visualstudio.com/Buying%20Catalogue/_workitems/edit/8092

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Catalogue Solutions Section Update.feature
```diff
@@ -0,0 +1,49 @@
...
+Scenario: 1. Update an orders catalogue solutions
+    When the user makes a request to update the order catalogue solutions section with the ID C000014-01
+    Then a response with status code 204 is returned
+    And the order with ID C000014-01 has catalogue solutions viewed set to true
+    And the order with orderId C000014-01 is updated in the database with data
+        | LastUpdatedBy                        | LastUpdatedByName |
+        | 7b195137-6a59-4854-b118-62b39a3101ef | Bob Smith         |
+    And the order with orderId C000014-01 has LastUpdated time present and it is the current time
+
+@4631
```
RuthBerry96, line 22: Should this be @7840?

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Builders/OrderBuilder.cs
```diff
@@ -1,131 +1,172 @@
-﻿using System;
-using NHSD.BuyingCatalogue.Ordering.Domain;
+﻿﻿using System;
+ using System.Collections.Generic;
```
RuthBerry96, line 2: got a rouge space here

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Catalogue Solutions Section Update.feature
sam-roberts-bjss: I think story 4631 fits with these tests. The change is in this PR because I renamed the file. I will leave these tags as they are.

PR #118, opened 2020-07-07
Improve the encapsulation of creating an order

sam-roberts-bjss:


src/NHSD.BuyingCatalogue.Ordering.Domain/OrderStatus.cs
```diff
@@ -1,9 +1,8 @@
 ﻿namespace NHSD.BuyingCatalogue.Ordering.Domain
 {
-    public sealed class OrderStatus
+    public enum OrderStatus : int
```
RuthBerry96, line 3: I dont think you have to do :int, as it does it as defualt

sam-roberts-bjss: 👍

PR #122, opened 2020-07-08
8119: Service Recipient Name Fix

RuthBerry96:
Story: [AB#8119](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8119)
Task: [AB#8249](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8249)

### DIFF
--- a/src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/CatalogueSolutionsController.cs
+++ b/src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/CatalogueSolutionsController.cs
@@ -88,14 +88,18 @@ namespace NHSD.BuyingCatalogue.Ordering.Api.Controllers
             if (orderItem is null)
                 return NotFound();
 
+            var serviceRecipients = order.ServiceRecipients;
+
             return new GetCatalogueSolutionOrderItemModel
             {
                 ServiceRecipient = new ServiceRecipientModel
                 {
-                    OdsCode = orderItem.OdsCode
+                    OdsCode = orderItem.OdsCode,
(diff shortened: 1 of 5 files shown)


src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/CatalogueSolutionsController.cs
```diff
@@ -88,14 +88,18 @@ public async Task<ActionResult<GetCatalogueSolutionOrderItemModel>> GetOrderItem
...
+            var serviceRecipients = order.ServiceRecipients;
+
             return new GetCatalogueSolutionOrderItemModel
             {
                 ServiceRecipient = new ServiceRecipientModel
                 {
-                    OdsCode = orderItem.OdsCode
+                    OdsCode = orderItem.OdsCode,
+                    Name = serviceRecipients.FirstOrDefault(serviceRecipient => string.Equals(orderItem.OdsCode,
+                        serviceRecipient.OdsCode, StringComparison.Ordinal))?.Name
```
CallumBradburyBJSS, line 99: I think this should be InvariantCulture or OrdinalIgnoreCase, not just Ordinal. Ordinal will treat case sensitivity differently... is that what we want?

RuthBerry96: This is using the same ones as other, so might be a tech debt issue, as we should then change the others ect. Or I can do it in my next PR, as I'm doing another service recipient name then

PR #133, opened 2020-07-14
Feature/Merge development 14072020

sam-roberts-bjss:
Activity: [AB#8123](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8123)
Task: [AB#8148](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8148)


.editorconfig
```diff
@@ -52,7 +52,7 @@ indent_size = 2
 
 # Feature files
 [*.feature]
-end_of_line = lf
+end_of_line = crlf
 insert_final_newline = true
```
SimonLLawrence, line 56: Are we going back to just good old crlf ?

sam-roberts-bjss: Thought try something new because we are still getting inconsistent line endings 😢

SimonLLawrence: ok

PR #139, opened 2020-07-15
Feature/7580 Add get order item endpoint

sam-roberts-bjss:
[AB#7580](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7580)
[AB#8378](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8378)


src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrderItemsController.cs
```diff
@@ -98,24 +98,36 @@ public async Task<ActionResult<IEnumerable<GetOrderItemModel>>> GetAllAsync(stri
...
-                    },
-                    CataloguePriceType = orderItem.CataloguePriceType.Name,
-                    CatalogueItemType = orderItem.CatalogueItemType.Name,
-                    CatalogueItemName = orderItem.CatalogueItemName,
-                    CatalogueItemId = orderItem.CatalogueItemId,
-                }).ToList();
+                .Select(orderItem => new GetOrderItemModel(orderItem, serviceRecipientDictionary[orderItem.OdsCode.ToUpperInvariant()])).ToList();
+        }
+
+        [HttpGet("{orderItemId}")]
```
RuthBerry96, line 108: Will it not be better to match the other routes, and have a seperate [Route] containing the orderItemId parameter rather than adding it to the get?

sam-roberts-bjss: Yes that makes sense. I will add the route attribute.

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Requests/GetOrderItemRequest.cs
```diff
@@ -0,0 +1,38 @@
...
+    internal sealed class GetOrderItemRequest
+    {
+        private readonly Request _request;
+        private readonly string _getOrderItemUrl;
+
+        public GetOrderItemRequest(
+            Request request,
+            string orderingApiBaseAddress,
+            string orderId,
+            int? orderItemId)
```
RuthBerry96, line 17: Can the orderItemId be of type int? 
In the controller its passed in as type int (unless I'm missing something)

sam-roberts-bjss: Yeah I was attempting to use this as a "not found" test. In the end I went with another step, which still not sure of but yeah will put this back to an int.

PR #144, opened 2020-07-20
5115: Update GET order summary to include associated

RuthBerry96:
Story: [AB#5115](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/5115)
Task: [AB#8529](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8529)


tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Controllers/OrdersControllerTests.cs
```diff
@@ -308,6 +309,46 @@ public async Task GetOrderSummaryAsync_CatalogueSolutionCount_ReturnsCountOfTwo(
             actual.Should().BeEquivalentTo(expected);
         }
 
+        [TestCase]
+        public async Task GetOrderSummaryAsync_AssociatedServiceCount_ReturnsCountOfTwo()
```
CallumBradburyBJSS, line 313: A suggestion: not required, but it'd be good to add a non associated service into the list to ensure it's filtering

RuthBerry96: 👍 good suggestion. Will add it in

PR #147, opened 2020-07-21
4832: Added support for the creation of Asssocated Service order items.

SimonLLawrence:
Story [AB#4832](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/4832)
Task [AB#8446](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8446)


tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Requests/CreateOrderItemBaseRequest.cs
```diff
@@ -60,6 +65,10 @@ public void SetPayload(string key)
         }
         public void AssertPayload(OrderItemEntity actual)
         {
+            var timeUnit = Payload.CatalogueItemType == CatalogueItemType.AssociatedService
```
stuart-bass-bjss, line 68: Why check the type and set, and not just check string.isnullorwhitespace(Payload.TimeUnitName)?

PR #153, opened 2020-07-23
5436: Add funding source flag to db

RuthBerry96:
Story: [AB#5436](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/5436)
Task: [AB#8654](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8654)


src/NHSD.BuyingCatalogue.Ordering.OrderingDatabase/Tables/Order.sql
```diff
@@ -15,8 +15,9 @@
     SupplierAddressId INT NULL,
     SupplierContactId INT NULL,
     CatalogueSolutionsViewed BIT DEFAULT 0,
-    AdditionalServicesViewed BIT DEFAULT 0,    
+    AdditionalServicesViewed BIT DEFAULT 0,
     AssociatedServicesViewed BIT DEFAULT 0,
+    FundingSourceViewed BIT DEFAULT 0,
```
stuart-bass-bjss, line 20: Funding source is more like commencement date. We don't have logic around when it's viewed, they have to choose yes or no.
I think it should be a nullable bit field, maybe 'FundingSourceOnlyGMS' or similar to reflect the question.

PR #166, opened 2020-07-31
8852: Added delete order endpoint

CallumBradburyBJSS:
Story: [AB#5287](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/5287)
Task: [AB#8852](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8852)


src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrdersController.cs
```diff
@@ -226,5 +226,31 @@ public async Task<ActionResult<CreateOrderResponseModel>> CreateOrderAsync([From
...
+        [HttpDelete]
+        [Route("{orderId}")]
+        [Authorize(Policy = PolicyName.CanManageOrders)]
+        public async Task<ActionResult> DeleteOrderAsync(string orderId)
+        {
+            var order = await _orderRepository.GetOrderByIdAsync(orderId);
+
+            if (order is null || order.IsDeleted)
+            {
+                return NotFound();
```
sam-roberts-bjss, line 240: I would like to suggest here that if the order has already been deleted the request was successful anyway. So I think it would be better to return NoContent, what do you think? I believe there tends to be a bit of debate around this subject in REST terms and that the request should be idempotent, this way the client doesn't have to cater for the NotFound case & it can be called multiple times & will result in the same response.

src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrdersController.cs
```diff
@@ -226,5 +226,31 @@ public async Task<ActionResult<CreateOrderResponseModel>> CreateOrderAsync([From
...
         }
+
+        [HttpDelete]
+        [Route("{orderId}")]
+        [Authorize(Policy = PolicyName.CanManageOrders)]
+        public async Task<ActionResult> DeleteOrderAsync(string orderId)
+        {
+            var order = await _orderRepository.GetOrderByIdAsync(orderId);
+
+            if (order is null || order.IsDeleted)
```
sam-roberts-bjss, line 238: I don't think the `order.IsDeleted` is really needed here especially after the work @SimonLLawrence is doing with the global query filter. Suppose it could be classed as belt & braces, so up to you if feel it could be removed.

src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrdersController.cs
```diff
@@ -226,5 +226,31 @@ public async Task<ActionResult<CreateOrderResponseModel>> CreateOrderAsync([From
...
+                return NotFound();
+            }
+
+            var primaryOrganisationId = User.GetPrimaryOrganisationId();
+            if (primaryOrganisationId != order.OrganisationId)
+            {
+                return Forbid();
+            }
+
+            order.IsDeleted = true;
```
sam-roberts-bjss, line 249: I would usually tend to prefer the use of encapsulation (i.e. Method on the order, something like `MarkForDeletion()`) here as it could specify & wrap up the setting of the last updated by information. This could also include any events we want to raise from the deleting of an order or the setting of the deleted date. However there is no requirement for this just yet. Something to consider, don't necessarily have to change now.

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Order Delete.feature
```diff
@@ -0,0 +1,61 @@
...
+        | OrderId    | Description   | OrganisationId                       | OrganisationName | OrganisationOdsCode | OrganisationContactEmail | SupplierAddressPostcode | SupplierContactEmail    |
+        | C000014-01 | A Description | 4af62b99-638c-4247-875e-965239cd0c48 | Hampshire CC     | 432432              | Fred.robinson@email.com  | LS 1 3AP                | Fred.robinson@email.com |
+        And the user is logged in with the Buyer role for organisation 4af62b99-638c-4247-875e-965239cd0c48
+
+Scenario: 1. Delete an existing order
+    Given the user creates a request to delete the order with ID 'C000014-01'
+    When the user sends the delete order request
+    Then a response with status code 204 is returned
+    And the expected order is deleted
+    And the order with orderId C000014-01 is updated in the database with data
```
sam-roberts-bjss, line 17: Would it be better to move this last assertion to another scenario as it looks like a secondary piece of behaviour? Hopefully makes the test easier to read & understand as you are testing one specific piece of change.

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Order Delete.feature
```diff
@@ -0,0 +1,61 @@
...
+
+@5287
+Scenario: 4. If a user is not authorised then they cannot delete an order
+    Given no user is logged in
+    Given the user creates a request to delete the order with ID 'C000014-01'
+    When the user sends the delete order request
+    Then a response with status code 401 is returned
+
+@5287
+Scenario: 5. A non buyer user cannot update an order
```
sam-roberts-bjss, line 43: update an order -> delete an order?

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Order Delete.feature
```diff
@@ -0,0 +1,61 @@
...
+    Given the user is logged in with the Authority role for organisation 4af62b99-638c-4247-875e-965239cd0c48
+    Given the user creates a request to delete the order with ID 'C000014-01'
+    When the user sends the delete order request
+    Then a response with status code 403 is returned
+
+@5287
+Scenario: 6. A buyer user cannot update an order for an organisation they don't belong to
```
sam-roberts-bjss, line 50: update -> delete

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Requests/DeleteOrderRequest.cs
```diff
@@ -0,0 +1,34 @@
...
+
+            _deleteOrderUrl =
+                $"{orderingApiBaseAddress}/api/v1/orders/{orderId}";
+        }
+
+        public string OrderId { get; }
+
+        public async Task<Response> ExecuteAsync()
+        {
+            var response = await _request.DeleteAsync(_deleteOrderUrl);
```
sam-roberts-bjss, line 30: Very minor, could just return rather than introduce a variable.

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Controllers/OrdersControllerTests.cs
```diff
@@ -460,6 +460,50 @@ async Task<ActionResult<CreateOrderResponseModel>> CreateOrder()
...
         }
 
+        [Test]
+        public async Task DeleteOrderAsync_DeleteOrderSuccessful_ReturnsNoContent()
+        {
+            var context = OrdersControllerTestContext.Setup();
+
+            var response = await context.OrdersController.DeleteOrderAsync("Order");
+            response.Should().BeOfType<NoContentResult>();
+            context.Order.IsDeleted.Should().BeTrue();
```
sam-roberts-bjss, line 470: Would also like to suggest splitting these assertions out as well, similar to a previous comment.

src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrdersController.cs
CallumBradburyBJSS: I don't think that's a particularly cool idea. If we ended up switching to a hard-delete for some reason, we'd then have a change in behaviour which would be mega-weird

src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrdersController.cs
CallumBradburyBJSS: As that work isn't in yet, I'll leave it in. Once Simon's stuffs in we could remove it though for sure

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Order Delete.feature
CallumBradburyBJSS: roger roger, good call

src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrdersController.cs
CallumBradburyBJSS: Yeah I think if we did add further behaviour based on deletion that'd be a good idea, for now I like the simplicity

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Order Delete.feature
CallumBradburyBJSS: I really thought I caught all those :(

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Requests/DeleteOrderRequest.cs
CallumBradburyBJSS: ah yes I meant to do this! 👍

src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrdersController.cs
sam-roberts-bjss: I'm not sure how it would change with a hard delete. It would result in the same response (200 / 204). The record no longer exists therefore it has been deleted. This discussion does come up a lot from various "REST" talks, maybe a second opinion would help?

CallumBradburyBJSS: So if we try and delete an order that has never existed, we would get NoContent? I'm not sure that makes sense. Let's get that 2nd opinion! @bjss-chris

bjss-chris: Hmm, thanks @CallumBradburyBJSS! This is a slightly tricky one. However, I think I'm leaning towards a 404 not found response.

In this case, I'm not sure the returned status code really affects the operation's idempotency. The order can only be deleted once and assuming that the implementation doesn't randomly start picking off other orders (I've not checked but I'm sure it won't!), i.e. it does nothing, then we can probably argue that the operation is idempotent. I guess it may depend on your definition of idempotent but I tend to think about state when considering whether an operation is idempotent.

[RFC 7321](https://tools.ietf.org/html/rfc7231#section-4.3.5) also states that

> If a DELETE method is successfully applied, the origin server SHOULD send... a 204 (No Content) status code if the action has been enacted and no further information is to be supplied, or a 200 (OK) status code if the action has been enacted and the response message includes a representation describing the status.

If we're not actually deleting anything have we enacted the action? My initial thought is: probably not.

However, I can also understand some of the arguments for returning the same response regardless. For the reasons I've tried to outline above, I think I would initially be tempted to go with the 404 though.

One of the issues here is, of course, the lack of an agreed/adopted standard for RESTful APIs. If one was agreed it would hopefully make some of these choices easier.

Whichever status code we decide to implement now it shouldn't be a huge amount of work to swap over later should it (if we had a reason to do so)?

Does that help at all? (I suspect not!)

CallumBradburyBJSS: Cheers Chris! I think if the FE accidentally sent the wrong ID, and it was for a non-existent order, it's better for them to be told the order doesn't exist, so they know they messed up, rather than being told everything was deleted successfully and assuming all is well.

sam-roberts-bjss: All good points 👍  We can go with the 404 then.

PR #167, opened 2020-07-31
Feature/5145 add order status to get order summary

sam-roberts-bjss:
Story: [AB#5145](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/5145)
Task: [AB#8911](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8911)

### DIFF
--- a/tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Builders/OrderSummaryModelBuilder.cs
+++ b/tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Builders/OrderSummaryModelBuilder.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using NHSD.BuyingCatalogue.Ordering.Api.Models.Summary;
+using NHSD.BuyingCatalogue.Ordering.Domain;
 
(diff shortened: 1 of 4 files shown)


tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Builders/OrderSummaryModelBuilder.cs
```diff
@@ -10,13 +11,15 @@ internal sealed class OrderSummaryModelBuilder
         private readonly string _description;
         private Guid _organisationId;
         private IEnumerable<SectionModel> _sections;
+        private readonly string _status;
```
CallumBradburyBJSS, line 14: Are there no unit tests to go along with this change?

sam-roberts-bjss: Yeah I looked at this. Quite a few tests were failing but by adding the default value they then passed. So I checked there were tests covering this functionality, but didn't add a specific test as we currently don't have the ability to change that status yet as per a user case. Don't mind either way, would you like me to add a specific test?

PR #168, opened 2020-07-31
Feature/5145 Change order status names

sam-roberts-bjss:
Story: [AB#5145](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/5145)
Task: [AB#8922](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8922)


src/NHSD.BuyingCatalogue.Ordering.OrderingDatabase/Deployment/CreateOrderStatuses.sql
```diff
@@ -1,10 +1,11 @@
...
 
 IF NOT EXISTS (
   SELECT *
   FROM dbo.OrderStatus
-  WHERE [Name] IN (@submitted, @unsubmitted))
+  WHERE [Name] IN (@complete, @incomplete))
 BEGIN
-	INSERT INTO dbo.OrderStatus ([Name]) VALUES (@submitted), (@unsubmitted)
+	INSERT INTO dbo.OrderStatus (OrderStatusId, [Name]) VALUES (1, @complete);
+    INSERT INTO dbo.OrderStatus (OrderStatusId, [Name]) VALUES (2, @incomplete);
```
bjss-chris, line 10: This could (and should, in my opinion) be a single `INSERT` statement.

I think there's also a mix of tabs and spaces here.

A bit of an aside, as I can see this is just an edit, but is there any reason why all the other post-deployment files (other than `CreateTables.sql` although I'm not sure why we still have that)  follow the format `Insert[Thing]` but this one is `Create[Thing]`?

src/NHSD.BuyingCatalogue.Ordering.OrderingDatabase/Tables/OrderStatus.sql
```diff
@@ -1,7 +1,7 @@
 ﻿CREATE TABLE dbo.OrderStatus
 (
-    OrderStatusId INT IDENTITY(1,1), 
-    [Name] NVARCHAR(30) NULL,
+    OrderStatusId INT NOT NULL,
```
bjss-chris, line 3: What's the reason for changing this from being an `IDENTITY` column?

src/NHSD.BuyingCatalogue.Ordering.OrderingDatabase/Deployment/CreateOrderStatuses.sql
sam-roberts-bjss: I can fix those 👍

src/NHSD.BuyingCatalogue.Ordering.OrderingDatabase/Tables/OrderStatus.sql
sam-roberts-bjss: I felt like this ensures more control over the ID of each status as these values are represented in the Domain layer, wouldn't want them to get out of step e.g. change in the order of insert. Also makes it clear what the ID is & I didn't have to run the select to find out.

bjss-chris: Admittedly, the `IDENTITY` here probably didn't really add much as these values will only ever be changed/added to as part of a release.

PR #181, opened 2020-08-13
8904: Add date completed

RuthBerry96:
Story: [AB#8904](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/8904)
Task: [AB#9346](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9346)

### DIFF
--- a/src/NHSD.BuyingCatalogue.Ordering.Domain/Order.cs
+++ b/src/NHSD.BuyingCatalogue.Ordering.Domain/Order.cs
@@ -9,6 +9,10 @@ namespace NHSD.BuyingCatalogue.Ordering.Domain
 {
     public sealed class Order
     {
+#pragma warning disable 649
+        private DateTime? _completed;
+#pragma warning restore 649
+
         private readonly List<OrderItem> _orderItems = new List<OrderItem>();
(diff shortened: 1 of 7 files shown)


src/NHSD.BuyingCatalogue.Ordering.Domain/Order.cs
```diff
@@ -42,7 +42,7 @@ private Order(OrderDescription orderDescription, Guid organisationId) : this()
 
         public DateTime Created { get; set; }
 
-        public DateTime? Completed { get; private set; }
+        public DateTime? Completed { get; set; }
```
sam-roberts-bjss, line 45: As discussed on the phone, this unfortunately is undoing the encapsulation of the Completed field. Really we only want to set this if the order has been marked as completed in the `Complete` method. This hopefully makes it easier to follow as there is only one place this value can be set. So need to use one of the methods e.g. Reflection to set this as part of a test. We already have an example of this in the `OrderItem` class for the `Created` property.

PR #182, opened 2020-08-17
Feature/3928 Send an email when an order is completed

sam-roberts-bjss:
Story: [AB#3928](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/3928)
Task: [AB#9298](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9298)


src/NHSD.BuyingCatalogue.Ordering.Api/Startup.cs
```diff
@@ -46,33 +50,50 @@ public void ConfigureServices(IServiceCollection services)
             var requireHttps = _configuration.GetValue<bool>("RequireHttps");
             var allowInvalidCertificate = _configuration.GetValue<bool>("AllowInvalidCertificate");
             var bypassIdentity = _configuration.GetValue<bool>("BypassIdentity");
-
+;
```
RuthBerry96, line 53: Is this a rouge semicolon?

assoomro, line 53: did this sneak in?

src/NHSD.BuyingCatalogue.Smtp/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM maildev/maildev
```
assoomro, line 1: We're already building and pushing an image of this (nhsd/buying-catalogue/smtp-server in the acr) - should we perhaps re-use this instead of duplicating it here?

src/NHSD.BuyingCatalogue.Smtp/stunnel.conf
```diff
@@ -0,0 +1,6 @@
+[smtp]
```
assoomro, line 1: Same as above, we could get rid of the whole .Smtp directory

src/NHSD.BuyingCatalogue.Smtp/Dockerfile
sam-roberts-bjss: Interested in getting your thoughts on this... How will it work?

src/NHSD.BuyingCatalogue.Smtp/stunnel.conf
sam-roberts-bjss: > Good work, few things to address.
> 
> Also - I imagine we'd need to amend our chart to accommodate for bringing the smtp server in, can we please do it as part of this PR?

I've created a separate task for this [9450](https://buyingcatalog.visualstudio.com/Buying%20Catalogue/_workitems/edit/9450). Put a message in the platform channel. Would be good to work out whether or not some of the settings can be reused.

assoomro: Ah, sorry, didn't see your broke this off into its own task, my bad. As far as I know most, if not all, of the smtp settings will be reusable.

azure-pipelines.yml
```diff
@@ -30,37 +31,37 @@ jobs:
...
-      ArtifactName: database
-
-  - task: colinsalmcorner.colinsalmcorner-buildtasks.dacpac-report-task.DacPacReport@1
-    displayName: 'DacPac Schema Compare NHSD.BuyingCatalogue.Ordering.OrderingDatabase.dacpac'
-    inputs:
-      dropName: database
-      dacpacName: NHSD.BuyingCatalogue.Ordering.OrderingDatabase
-      targetDacPacPath: '$(build.artifactstagingdirectory)/database'
-    enabled: false
+# - job: build
```
CallumBradburyBJSS, line 34: A whole lot of stuff commented out in this file, can I get a quick summary?

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Controllers/OrdersControllerTests.cs
```diff
@@ -633,15 +614,12 @@ public async Task UpdateStatusAsync_OrderIsComplete_LastUpdatedIsUpdated()
 
             await context.OrdersController.UpdateStatusAsync("Order", context.CompleteOrderStatusModel);
 
-            context.Order.LastUpdatedBy.Should().Be(context.NameIdentity);
-            context.Order.LastUpdatedByName.Should().Be(context.Name);
-            context.Order.LastUpdated.Should().NotBe(DateTime.MinValue);
+            context.CompleteOrderServiceMock
+                .Verify(x => x.CompleteAsync(It.IsNotNull<CompleteOrderRequest>()), Times.Once);
```
CallumBradburyBJSS, line 618: Could update this to instead of doing It.IsNotNull, use It.Is<CompleteOrderRequest>(req => req.Order == context.Order), I think, which would be a bit more specific

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Requests/CompleteOrderRequestTests.cs
```diff
@@ -0,0 +1,17 @@
...
+using NUnit.Framework;
+
+namespace NHSD.BuyingCatalogue.Ordering.Api.UnitTests.Requests
+{
+    [TestFixture]
+    [Parallelizable(ParallelScope.All)]
+    internal sealed class CompleteOrderRequestTests
+    {
+        [Test]
+        public void Constructor_NullOrder_ThrowsArgumentNullException()
```
CallumBradburyBJSS, line 12: A test for the happy path is probably worth adding too, making sure .Order is what's passed in

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Services/CompleteOrderService.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        [Test]
+        public async Task CompleteAsync_OrderRepository_CalledOnce()
+        {
+            var context = CompleteOrderServiceTestContext.Setup();
+
+            await context.CompleteOrderService.CompleteAsync(context.CompleteOrderRequest);
+
+            context.OrderRepositoryMock.Verify(orderRepository => 
+                orderRepository.UpdateOrderAsync(It.IsNotNull<Order>()), Times.Once);
```
CallumBradburyBJSS, line 93: Can update this verify to ensure it's updating the correct order

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Services/CompleteOrderService.cs
```diff
@@ -0,0 +1,170 @@
...
+            var context = CompleteOrderServiceTestContext.Setup();
+
+            var actual = await context.CompleteOrderService.CompleteAsync(
+                new CompleteOrderRequest(OrderBuilder.Create().Build()));
+
+            actual.Should().Be(Result.Failure(new ErrorDetails("OrderNotComplete")));
+        }
+
+        [Test]
+        public async Task CompleteAsync_IdentityService_CalledTwice()
```
CallumBradburyBJSS, line 72: A test to verify that the Order is being updated with the data received from these calls would be useful

azure-pipelines.yml
sam-roberts-bjss: Apologies, this was me testing out the pipeline to try and getting it working with the NuGet.config file. Turns out if I move it to the root of the project it works fine. So have made that change instead.

src/NHSD.BuyingCatalogue.Smtp/Dockerfile
sam-roberts-bjss: Probably need to discuss this with @stuart-bass-bjss as I am not keen having to download an image, should be able to work on your project offline. Maybe I am clinging on to a requirement not needed anymore in a connected world. But I think you highlight a good point. We need a way of sharing this & I am not currently sure of the best approach of this.

CallumBradburyBJSS (approved):

PR #190, opened 2020-08-28
Feature/9283 populate csv appropriate columns

RuthBerry96:
Story: [AB#9283](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9283)
Task: [AB#9642](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9642)

### DIFF
--- a/src/NHSD.BuyingCatalogue.Ordering.Api/Services/CreatePurchasingDocument/AttachmentCsvWriter.cs
+++ b/src/NHSD.BuyingCatalogue.Ordering.Api/Services/CreatePurchasingDocument/AttachmentCsvWriter.cs
@@ -1,0 +1,29 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.Threading.Tasks;
+using CsvHelper;
+using CsvHelper.Configuration;
+using CsvHelper.TypeConversion;
+
+namespace NHSD.BuyingCatalogue.Ordering.Api.Services.CreatePurchasingDocument
+{
+    public sealed class AttachmentCsvWriter<T, T2> : IAttachmentCsvWriter<T> where T2 : ClassMap<T>
+    {
+        public async Task WriteRecordsAsync(Stream stream, IEnumerable<T> records)
+        {
+            if (stream is null)
+                throw new ArgumentNullException(nameof(stream));
+
+            await using var streamWriter = new StreamWriter(stream, leaveOpen: true);
+            await using var csvWriter = new CsvWriter(streamWriter, CultureInfo.CurrentCulture);
+
+            var options = new TypeConverterOptions { Formats = new[] { "dd/MM/yyyy" } };
+            csvWriter.Configuration.TypeConverterOptionsCache.AddOptions<DateTime?>(options);
+
+            csvWriter.Configuration.RegisterClassMap<T2>();
+            await csvWriter.WriteRecordsAsync(records);
+        }
+    }
+}
--- a/src/NHSD.BuyingCatalogue.Ordering.Api/Services/CreatePurchasingDocument/CsvStreamStreamWriter.cs
+++ b/src/NHSD.BuyingCatalogue.Ordering.Api/Services/CreatePurchasingDocument/CsvStreamStreamWriter.cs
@@ -9,7 +9,7 @@ using CsvHelper.TypeConversion;
 
 namespace NHSD.BuyingCatalogue.Ordering.Api.Services.CreatePurchasingDocument
 {
-    internal sealed class CsvStreamWriter<TEntity, TClassMap> : IAttachmentCsvWriter<TEntity> where TClassMap : ClassMap<TEntity>
+    internal sealed class CsvStreamStreamWriter<TEntity, TClassMap> : ICsvStreamWriter<TEntity> where TClassMap : ClassMap<TEntity>
     {
         public async Task WriteRecordsAsync(Stream stream, IEnumerable<TEntity> records)
         {
--- a/src/NHSD.BuyingCatalogue.Ordering.Api/Services/CreatePurchasingDocument/PatientNumbersPriceTypeMap.cs
+++ b/src/NHSD.BuyingCatalogue.Ordering.Api/Services/CreatePurchasingDocument/PatientNumbersPriceTypeMap.cs
@@ -2,7 +2,7 @@
 
 namespace NHSD.BuyingCatalogue.Ordering.Api.Services.CreatePurchasingDocument
 {
-    public sealed class PatientNumbersPriceTypeMap : ClassMap<PatientNumbersPriceType>
+    internal sealed class PatientNumbersPriceTypeMap : ClassMap<PatientNumbersPriceType>
     {
         public PatientNumbersPriceTypeMap()
         {
--- a/src/NHSD.BuyingCatalogue.Ordering.Api/Services/CreatePurchasingDocument/PatientNumbersPriceTypeMap.cs
+++ b/src/NHSD.BuyingCatalogue.Ordering.Api/Services/CreatePurchasingDocument/PatientNumbersPriceTypeMap.cs
@@ -1,0 +1,31 @@
+﻿using CsvHelper.Configuration;
+
+namespace NHSD.BuyingCatalogue.Ordering.Api.Services.CreatePurchasingDocument
+{
+    public sealed class PatientNumbersPriceTypeMap : ClassMap<PatientNumbersPriceType>
+    {
+        public PatientNumbersPriceTypeMap()
+        {
+            Map(x => x.CallOffAgreementId).Index(0).Name("Call Off Agreement ID");
+            Map(x => x.CallOffOrderingPartyId).Index(1).Name("Call Off Ordering Party ID");
+            Map(x => x.CallOffOrderingPartyName).Index(2).Name("Call Off Ordering Party Name");
+            Map(x => x.CallOffCommencementDate).Index(3).Name("Call Off Commencement Date");
+            Map(x => x.ServiceRecipientId).Index(4).Name("Service Recipient ID");
+            Map(x => x.ServiceRecipientName).Index(5).Name("Service Recipient Name");
+            Map(x => x.ServiceRecipientItemId).Index(6).Name("Service Recipient Item ID");
+            Map(x => x.SupplierId).Index(7).Name("Supplier ID");
+            Map(x => x.SupplierName).Index(8).Name("Supplier Name");
+            Map(x => x.ProductId).Index(9).Name("Product ID");
+            Map(x => x.ProductName).Index(10).Name("Product Name");
+            Map(x => x.ProductType).Index(11).Name("Product Type");
+            Map(x => x.QuantityOrdered).Index(12).Name("Quantity Ordered");
+            Map(x => x.UnitOfOrder).Index(13).Name("Unit of Order");
+            Map(x => x.Price).Index(14).Name("Price");
+            Map(x => x.FundingType).Index(15).Name("Funding Type");
+            Map(x => x.M1Planned).Index(16).Name("M1 planned (Delivery Date)");
+            Map(x => x.ActualM1Date).Index(17).Name("Actual M1 date");
+            Map(x => x.BuyerVerificationDate).Index(18).Name("Buyer verification date (M2)");
+            Map(x => x.CeaseDate).Index(19).Name("Cease Date");
+        }
+    }
+}
--- a/tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Steps/EmailSteps.cs
+++ b/tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Steps/EmailSteps.cs
@@ -1,12 +1,15 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using FluentAssertions;
-using System.Net.Mime;
-using System.Text;
+using Microsoft.VisualBasic.FileIO;
 using NHSD.BuyingCatalogue.EmailClient.IntegrationTesting.Data;
 using NHSD.BuyingCatalogue.EmailClient.IntegrationTesting.Drivers;
+using NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests.Steps.Support;
+using NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests.Support;
+using NHSD.BuyingCatalouge.Ordering.Api.Testing.Data.Entities;
 using TechTalk.SpecFlow;
 using TechTalk.SpecFlow.Assist;
 using static System.String;
@@ -17,10 +20,14 @@ namespace NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests.Steps
     internal sealed class EmailSteps
     {
         private readonly EmailServerDriver _emailServerDriver;
+        private readonly OrderContext _orderContext;
+        private readonly ScenarioContext _scenarioContext;
 
-        public EmailSteps(EmailServerDriver emailServerDriver)
+        public EmailSteps(EmailServerDriver emailServerDriver, OrderContext orderContext, ScenarioContext scenarioContext)
         {
             _emailServerDriver = emailServerDriver ?? throw new ArgumentNullException(nameof(emailServerDriver));
+            _orderContext = orderContext ?? throw new ArgumentNullException(nameof(orderContext));
+            _scenarioContext = scenarioContext ?? throw new ArgumentNullException(nameof(scenarioContext));
         }
 
         [Then(@"only one email is sent")]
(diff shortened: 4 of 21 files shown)


src/NHSD.BuyingCatalogue.Ordering.Api/Services/CreatePurchasingDocument/AttachmentCsvWriter.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Globalization;
+using System.IO;
+using System.Threading.Tasks;
+using CsvHelper;
+using CsvHelper.Configuration;
+using CsvHelper.TypeConversion;
+
+namespace NHSD.BuyingCatalogue.Ordering.Api.Services.CreatePurchasingDocument
+{
+    public sealed class AttachmentCsvWriter<T, T2> : IAttachmentCsvWriter<T> where T2 : ClassMap<T>
```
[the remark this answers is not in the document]
sam-roberts-bjss: Could the T's get a little more context, just to help with understanding on first review. e.g T = TEntity / TRecord & T2 = TConfig / TClassMap

src/NHSD.BuyingCatalogue.Ordering.Api/Services/CreatePurchasingDocument/AttachmentCsvWriter.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        public async Task WriteRecordsAsync(Stream stream, IEnumerable<T> records)
```
sam-roberts-bjss, line 14: Do we want to allow `NULL` records?

src/NHSD.BuyingCatalogue.Ordering.Api/Services/CreatePurchasingDocument/PatientNumbersPriceTypeMap.cs
```diff
@@ -0,0 +1,31 @@
+﻿using CsvHelper.Configuration;
+
+namespace NHSD.BuyingCatalogue.Ordering.Api.Services.CreatePurchasingDocument
+{
+    public sealed class PatientNumbersPriceTypeMap : ClassMap<PatientNumbersPriceType>
```
sam-roberts-bjss, line 5: internal?

src/NHSD.BuyingCatalogue.Ordering.Api/Services/CreatePurchasingDocument/PatientNumbersPriceTypeMap.cs
```diff
@@ -0,0 +1,31 @@
...
+            Map(x => x.UnitOfOrder).Index(13).Name("Unit of Order");
+            Map(x => x.Price).Index(14).Name("Price");
+            Map(x => x.FundingType).Index(15).Name("Funding Type");
+            Map(x => x.M1Planned).Index(16).Name("M1 planned (Delivery Date)");
+            Map(x => x.ActualM1Date).Index(17).Name("Actual M1 date");
+            Map(x => x.BuyerVerificationDate).Index(18).Name("Buyer verification date (M2)");
+            Map(x => x.CeaseDate).Index(19).Name("Cease Date");
+        }
+    }
+}
```
sam-roberts-bjss, line 31: New line to make Github happy?

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Steps/EmailSteps.cs
```diff
@@ -37,46 +44,116 @@ public async Task EmailIsNotSent()
...
+                "Funding Type",
+                "M1 planned (Delivery Date)",
+                "Actual M1 date",
+                "Buyer verification date (M2)",
+                "Cease Date"
+            };
+
+            var actual = (await _emailServerDriver.FindAllEmailsAsync()).First().Attachments.First();
+
+            var attachmentData = new MemoryStream(actual.AttachmentData.ToArray());
```
sam-roberts-bjss, line 93: Should the attachmentData stream be disposed?

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Steps/EmailSteps.cs
```diff
@@ -37,46 +44,116 @@ public async Task EmailIsNotSent()
...
-            
+
             var actual = (await _emailServerDriver.FindAllEmailsAsync()).First();
 
             actual.Should().BeEquivalentTo(expected, conf => conf.IncludingAllDeclaredProperties());
         }
 
+        [Then(@"the patient numbers price type attachment contains the correct information")]
+        public async Task ThenThePatientNumbersPriceTypeAttachmentContainsTheCorrectInformation()
+        {
```
sam-roberts-bjss, line 67: One suggestion here would be to refactor the code within the step to make it a bit more readable. It looks like there are two assertions here one for the header and one for the body of the attachment. To be honest if could use something like the "Request" / "Response" type construct we have used for some of the other steps e.g. `PurchaseDocumentEmailResponse` that could read the data out of the email & the step can then easily assert the information.

Also we will need some way of identifying the email attachment as won't be able to rely on the "first" when the next story is completed.

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Steps/EmailSteps.cs
```diff
@@ -37,46 +44,116 @@ public async Task EmailIsNotSent()
...
+                csvText.EndOfData.Should().BeFalse();
+
+                var row = csvText.ReadFields();
+
+                var data = CreatePatientNumbersTypeAttachmentData(order, orderItem, serviceRecipients);
+
+                row.Should().BeEquivalentTo(data);
+            }
+
+            csvText.EndOfData.Should().BeTrue();
```
sam-roberts-bjss, line 116: Rather than asserting each row of the email attachment, do you think it would be better to include one assert statement that checks the two complete lists? Meaning generate a list of string[] for the actual data and expected data. This would hopefully make the test easier to read & if there was more than one errors in the document the test failure result would contain all issues not use the first row to contain a mismatch.

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Steps/EmailSteps.cs
```diff
@@ -37,46 +44,116 @@ public async Task EmailIsNotSent()
...
             };
-            
+
             var actual = (await _emailServerDriver.FindAllEmailsAsync()).First();
 
             actual.Should().BeEquivalentTo(expected, conf => conf.IncludingAllDeclaredProperties());
         }
 
+        [Then(@"the patient numbers price type attachment contains the correct information")]
+        public async Task ThenThePatientNumbersPriceTypeAttachmentContainsTheCorrectInformation()
```
sam-roberts-bjss, line 66: The context of the step is very specific to the Update order status feature. Do you think it would be better if this step was placed in the `OrderStatusSteps` class? I believe this way you could use the details of the request and not have to store the orderid within the scenario context.

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Builders/Services/CreatePurchasingDocumentServiceBuilder.cs
```diff
@@ -0,0 +1,26 @@
+﻿using NHSD.BuyingCatalogue.Ordering.Api.Services.CreatePurchasingDocument;
+
+namespace NHSD.BuyingCatalogue.Ordering.Api.UnitTests.Builders.Services
+{
+    internal sealed class CreatePurchasingDocumentServiceBuilder
+    {
+        private IAttachmentCsvWriter<PatientNumbersPriceType> _patientNumbersCsvWriter;
+
+        public CreatePurchasingDocumentServiceBuilder()
```
sam-roberts-bjss, line 9: Private constructor?

src/NHSD.BuyingCatalogue.Ordering.Api/Services/CreatePurchasingDocument/AttachmentCsvWriter.cs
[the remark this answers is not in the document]
sam-roberts-bjss: The context of this class is to wrap the CsvWriter functionality. Just thinking of a name change, what do you think to `CsvStreamWriter`? Just because the scenario we are using it in at the moment is for an attachment, I think am I right in saying it could be used to create a file for example?

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Services/CreatePurchasingDocumentServiceTests.cs
```diff
@@ -13,32 +14,61 @@ namespace NHSD.BuyingCatalogue.Ordering.Api.UnitTests.Services
...
             Assert.ThrowsAsync<ArgumentNullException>(() =>
-                purchasingDocumentService.CreateDocumentAsync(
+                context.CreatePurchasingDocumentService.CreatePatientNumbersCsvAsync(
                     hasStream ? new MemoryStream() : null,
                     hasOrder ? OrderBuilder.Create().Build() : null));
         }
 
         [Test]
-        public async Task CreateDocumentAsync_DocumentIsCreated_StreamIsAsExpected()
+        public async Task CreatePatientNumbersCsvAsync_DocumentIsCreated_StreamIsAsExpected()
```
sam-roberts-bjss, line 41: Name of this test is unexpected. Am I right in thinking rather than checking the contents of the stream it now ensuring the AttachmentCsvWriter is called once?

sam-roberts-bjss (changes requested):
Put quite a few questions and suggestions on, see what you think?

src/NHSD.BuyingCatalogue.Ordering.Api/Services/CreatePurchasingDocument/AttachmentCsvWriter.cs
RuthBerry96: I thought about this. It doesnt specify? I guess not, as an order wouldnt be able to be complete if there were no records. So will add a null check in

src/NHSD.BuyingCatalogue.Ordering.Api/Services/CreatePurchasingDocument/PatientNumbersPriceTypeMap.cs
RuthBerry96: hmmm, it currently has a new line e.g. goes onto line 32, but will retry

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Steps/EmailSteps.cs
RuthBerry96: I guess so. The reason for implementing it like this, was being more explicit about what the error is. E.g. if the file still had another row in, then the endOfData would be false

RuthBerry96: *I still prefer this method tho

src/NHSD.BuyingCatalogue.Ordering.Api/Services/CreatePurchasingDocument/AttachmentCsvWriter.cs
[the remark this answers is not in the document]
RuthBerry96: Yep, you're right 😄

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Steps/EmailSteps.cs
RuthBerry96: ooo yes. Should probably use an using here and that should sort it out I think

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Steps/EmailSteps.cs
RuthBerry96: I think I was going to do a filename selector for the next story? I can split out the header and the content. But it makes sense its in the same section, as it is testing the contents of a single csv file (it's just bloated as the csvheader content is quite large

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Steps/OrderStatusSteps.cs
```diff
@@ -1,10 +1,17 @@
...
 using System.Threading.Tasks;
 using FluentAssertions;
+using Microsoft.VisualBasic.FileIO;
+using NHSD.BuyingCatalogue.EmailClient.IntegrationTesting.Drivers;
+using NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests.Support;
 using NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests.Utils;
 using NHSD.BuyingCatalouge.Ordering.Api.Testing.Data.Data;
 using NHSD.BuyingCatalouge.Ordering.Api.Testing.Data.Entities;
 using TechTalk.SpecFlow;
+using static System.String;
```
sam-roberts-bjss, line 14: I think this type of syntax was frowned upon in one of the coding standard meetings. Do you need to use a static using statement here?

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Services/CreatePurchasingDocumentServiceTests.cs
```diff
@@ -13,32 +14,63 @@ namespace NHSD.BuyingCatalogue.Ordering.Api.UnitTests.Services
...
             var order = OrderBuilder.Create().Build();
-            await purchasingDocumentService.CreateDocumentAsync(stream, OrderBuilder.Create().Build());
 
-            var expectedStreamContent = stream.ToArray();
-            var actualStreamContent = Encoding.UTF8.GetBytes($"Call off Party Id{Environment.NewLine}{order.OrderId}{Environment.NewLine}");
+            await context.CreatePurchasingDocumentService.CreatePatientNumbersCsvAsync(stream, order);
 
-            expectedStreamContent.Should().BeEquivalentTo(actualStreamContent);
+            context.PatientCsvWriterMock.Verify(
+                x => x.WriteRecordsAsync(It.IsAny<Stream>(), It.IsAny<IEnumerable<PatientNumbersPriceType>>()),
```
sam-roberts-bjss, line 50: Would it be possible to validate the mapping part of the `CreatePatientNumbersCsvAsync` method? So make sure the list of `PatientNumbersPriceType` is what you expect based on the supplied order.

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Services/CreatePurchasingDocumentServiceTests.cs
```diff
@@ -13,32 +14,63 @@ namespace NHSD.BuyingCatalogue.Ordering.Api.UnitTests.Services
...
+            PatientCsvWriterMock.Setup(x =>
+                x.WriteRecordsAsync(It.IsAny<Stream>(), It.IsAny<IEnumerable<PatientNumbersPriceType>>()));
+
+            CreatePurchasingDocumentService = CreatePurchasingDocumentServiceBuilder
+                .Create()
+                .WithPatientNumbersCsvWriter(PatientCsvWriterMock.Object)
+                .Build();
+        }
+
+        internal Mock<ICsvStreamWriter<PatientNumbersPriceType>> PatientCsvWriterMock;
```
sam-roberts-bjss, line 69: Is this a property or a member variable? Is it missing the `{ get; }` syntax?

src/NHSD.BuyingCatalogue.Ordering.Api/Services/CreatePurchasingDocument/CsvStreamStreamWriter.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Globalization;
+using System.IO;
+using System.Threading.Tasks;
+using CsvHelper;
+using CsvHelper.Configuration;
+using CsvHelper.TypeConversion;
+
+namespace NHSD.BuyingCatalogue.Ordering.Api.Services.CreatePurchasingDocument
+{
+    internal sealed class CsvStreamStreamWriter<TEntity, TClassMap> : ICsvStreamWriter<TEntity> where TClassMap : ClassMap<TEntity>
```
CallumBradburyBJSS, line 12: This is a really cool class, wow!

CallumBradburyBJSS (approved):

PR #204, opened 2020-09-22
Fix CSV output and encoding bugs

bjss-chris:
Also fixes a number of unit test and integration test issues.  In addition, there is also some clean-up in preparation for re-enabling of suppressed FxCop and StyleCop rules.

Bug: [AB#9771](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9771)
Task: [AB#9835](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9835)


PhysicsJon (approved):
Looks good

PR #274, opened 2021-02-10
Increase precision of price column to 4 decimal places in OrderItem table

RahulJha-mastek:
Story [AB#11143](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/11143)
Task [AB#11721](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/11721)

Price column in OrderItem table is update to 4 decimal places.


tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Order Update Status.feature
```diff
@@ -151,7 +151,7 @@ Scenario: When an order is complete, an the funding source is true, but the orde
...
-        | C000014-04            | OrgOds                     | NHS NORTHUMBERLAND CCG       | 07/05/2020                 | au                   | AU Test                | C000014-04-au-5           | 10101       | Supplier 1    | 100001-001 | Item 5       | Additional Service | 1                | per patient   | per month | per month         | 599.990 | 1          | Central      | 05/09/2021                 |                |                              |            |
+        | C000014-04            | OrgOds                     | NHS NORTHUMBERLAND CCG       | 07/05/2020                 | au                   | AU Test                | C000014-04-au-5           | 10101       | Supplier 1    | 100001-001 | Item 5       | Additional Service | 1                | per patient   | per month | per month         | 599.9900 | 1          | Central      | 05/09/2021                 |                |                              |            |
```
bjss-chris, line 154: I think it would make sense to use `599.9999` as the price here. I know the original was `599.990` but I think that might have been better as `599.999`. (This applies to all of the prices changed in this file.)

Can you also make sure that the vertical bars are all aligned? (This applies to all of the tables changed in this file.)

bjss-chris (changes requested):
There are just some to changes to make to the feature file but otherwise your changes look fine.

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Order Update Status.feature
RahulJha-mastek: I have updated data and also aligned table in feature file.

bjss-chris (approved):
Your last commit message should be in the imperative, so instead of 'Price value updated in feature file' it should be 'Update price value in feature file'.

I'll still approve your PR but try to keep an eye on the wording of your commit messages.

PR #298, opened 2021-03-04
Update domain objects

bjss-chris:

Unit test changes to follow in separate PR


bjss-chris:
The `LastUpdated` fields are still set but they're now set automatically whenever we update the DB (hence the removal of code that explicitly sets those values).

PhysicsJon (approved):

PR #307, opened 2021-03-05
Update extension methods and corresponding tests

bjss-chris:


PhysicsJon (approved):

src/NHSD.BuyingCatalogue.Ordering.Api/Extensions/AddressExtensions.cs
```diff
@@ -25,24 +24,22 @@ internal static AddressModel ToModel(this Address address)
             };
         }
 
-        internal static Address FromModel(this Address address, AddressModel model)
+        internal static Address ToDomain(this AddressModel model)
```
bjss-chris, line 27: This method probably ought to be in an `AddressModelExtensions` class... I'll split it out later.

PR #308, opened 2021-03-05
Improve model binding

bjss-chris:


src/NHSD.BuyingCatalogue.Ordering.Api/ModelBinding/OrderModelBinderProvider.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.AspNetCore.Mvc.ModelBinding.Binders;
+using NHSD.BuyingCatalogue.Ordering.Domain;
+
+namespace NHSD.BuyingCatalogue.Ordering.Api.ModelBinding
+{
+    internal sealed class OrderModelBinderProvider : IModelBinderProvider
```
bjss-chris, line 7: This is a real pain to test (mocking doesn't work so requires a number of supporting test classes). Therefore, I decided it wasn't worth the time at this point.

PhysicsJon (approved):

PR #311, opened 2021-03-05
Update controllers and tests

bjss-chris:
[AB#11053](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/11053)


bjss-chris:
We'll probably want to move the DB code out of the controllers at some point but that should be a fairly straight-forward refactor we can do later.

PhysicsJon (approved):
Not overly keen on the db in controller thing, but as we know it needs doing afterwards I'm not too fussed for now

PR #316, opened 2021-03-05
Fix order creation issue

bjss-chris:
Story: [AB#12939](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/12939)


src/NHSD.BuyingCatalogue.Ordering.Api/Controllers/OrdersController.cs
```diff
@@ -129,7 +129,7 @@ public async Task<IActionResult> CreateOrderAsync(CreateOrderModel model)
             return CreatedAtAction(
                 nameof(GetAsync).TrimAsync(),
                 new { callOffId = order.CallOffId.ToString() },
-                null);
+                new { orderId = order.CallOffId.ToString() });
```
bjss-andrew, line 132: is that right? callOffId and orderId are the same value?

bjss-chris: Yeah, it's due to bad use of domain language.

Outside of code it's always a call off ID but there's places in code (such as the order form) where it's still referred to as an order ID. It would be good to fix that but probably better to treat it as a tech debt task.

bjss-andrew (approved):

PR #372, opened 2021-04-06
Enable Complete Order button when Catalogue Solution is deleted and Associated Services remain on the order

prakashpawarnhs:
Story [AB#9039](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9039)  
Task [AB#13500](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13500)


src/NHSD.BuyingCatalogue.Ordering.Domain/Order.cs
```diff
@@ -168,7 +168,7 @@ public bool CanComplete()
             int associatedServicesCount = OrderItems.Count(o => o.CatalogueItem.CatalogueItemType.Equals(CatalogueItemType.AssociatedService));
 
             var solutionAndAssociatedServices = catalogueSolutionsCount > 0
-                && associatedServicesCount > 0;
+                || associatedServicesCount > 0;
```
bjss-chris, line 171: This condition is checking for the scenario where both a solution and an associated service exist so it needs to be an `AND` condition. Changing this condition to an `OR` makes it possible to complete the order in scenarios where it shouldn't be possible (please see my comments below on the integration tests). Therefore this change will need to be reverted.

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Order Summary Section Status.feature
```diff
@@ -181,7 +181,7 @@ Scenario: the order section status is set for an imcomplete order with one solut
         | 10001   | 1000-001        | ODS1    |
     When the user makes a request to retrieve the order summary with the ID 10001
     Then a response with status code 200 is returned
-    And the order Section Status is incomplete
+    And the order Section Status is complete
```
bjss-chris, line 184: This scenario is incorrect now. When there is a catalogue solution in the order, the order section status should only be complete once the additional services and associated services sections have also been viewed. Therefore this change will need to be reverted.

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Order Summary Section Status.feature
```diff
@@ -201,7 +201,7 @@ Scenario: the order section status is set for an incomplete order with one solut
         | 10001   | 1000-001-A01    | ODS2    |
     When the user makes a request to retrieve the order summary with the ID 10001
     Then a response with status code 200 is returned
-    And the order Section Status is incomplete
+    And the order Section Status is complete
```
bjss-chris, line 204: Similar to above this scenario is incorrect now. When there is a catalogue solution and an additional service in the order, the order section status should only be complete once the associated services section has also been viewed. Therefore this change will need to be reverted.

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Extensions/OrderExtensionsTests.cs
```diff
@@ -161,8 +161,8 @@ public static void IsSectionStatusCompleteComplete_NullOrder_ReturnsFalse()
         [CommonInlineAutoData(true, true, true, true, true, true, true)]
         [CommonInlineAutoData(true, true, true, true, false, true, true)]
         [CommonInlineAutoData(true, true, false, false, false, false, false)]
-        [CommonInlineAutoData(true, true, false, true, true, false, false)]
-        [CommonInlineAutoData(true, false, true, false, false, true, false)]
+        [CommonInlineAutoData(true, true, false, true, true, false, true)]
+        [CommonInlineAutoData(true, false, true, false, false, true, true)]
```
bjss-chris, line 165: Based on my comments about the scenarios above, these two test cases are also now incorrect so should be reverted.

bjss-chris (changes requested):
The wrong condition has been changed in `Order.CanComplete`. The condition that applies to this particular scenario is `noSolutionsAndAssociatedServices` on line 177.

The reason it fails is because there is an omission in the EF core query in `OrdersController.GetOrderSummaryAsync` – we should be including `OrderItemRecipients` but currently we aren't. Therefore the `serviceRecipientsCount` in `Order.CanComplete` is always `0`.

To fix the bug we'll need to add an extra line to the query:

```c#
.Include(o => o.OrderItems).ThenInclude(i => i.OrderItemRecipients)
```

after

```c#
.Include(o => o.OrderItems).ThenInclude(i => i.CatalogueItem)
```

I think we'll need to add an extra scenario to the integration tests to cover this particular case.

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Extensions/OrderExtensionsTests.cs
prakashpawarnhs: Revert is done

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Order Summary Section Status.feature
prakashpawarnhs: Revert is done

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Order Summary Section Status.feature
prakashpawarnhs: Revert is done

src/NHSD.BuyingCatalogue.Ordering.Domain/Order.cs
prakashpawarnhs: Revert is done and changes in OrderController file is pushed.

bjss-chris (approved):

PR #376, opened 2021-04-07
Update Orders API to return Recipient specific CostPerYear

RahulJha-mastek:
Story AB#131476
Task [AB#13653](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13653)


bjss-chris (changes requested):
Changes look fine but can you have a look to see if we can either update an existing scenario to cover this and add a new one (if necessary) in the Order Get Calculations feature?

RahulJha-mastek:
Added new scenario in feature file.

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Order Get Calculations.feature
```diff
@@ -76,3 +77,25 @@ Scenario: verify the price calculations for an order with multiple order items
...
+        | OrderId | CatalogueItemId | Price   | ProvisioningType   | PriceTimeUnit |
+        | 10001   | 10001-001       | 5       | Patient            | Year          |
+    And order item recipients exist
+        | OrderId | CatalogueItemId | OdsCode   | Quantity   |
+        | 10001   | 10001-001       | eu        | 10         |
+        | 10001   | 10001-001       | au        | 20         |
+    And the user creates a request to retrieve the details of an order by ID 10001
+    When the user sends the get order request
+    Then a response with status code 200 is returned
+    And the get order response contains recipient with <OdsCode> a yearly value of <RecipientTotal>
```
bjss-chris, line 95: I think this might be better if it was something like `And the get order response contains a recipient with <OdsCode> for catalogue item ID [ID-here] and a yearly value of <RecipientTotal>`.

We could then filter by catalogue ID.

What do you think?

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Responses/GetOrderResponse.cs
```diff
@@ -43,6 +43,14 @@ public void AssertOrderItemCost(decimal orderItemCost)
             actual.Should().Be(orderItemCost);
         }
 
+        public void AssertOrderItemRecipientCost(string odsCode, decimal orderItemCost)
+        {
+            var item = ReadOrderItems(ContentAsJson).First(i => i.CatalogueItemName is not null);
+            ExpectedServiceRecipient serviceRecipient = item.ServiceRecipients.First(i => i.OdsCode == odsCode);
```
bjss-chris, line 49: If we make the change suggested above we could select the item based on the item ID rather than testing the name.

I'm not sure we gain anything by testing that the item name is not null – the `Name` column has a `NOT NULL` constraint so a catalogue item name should never be null.

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Steps/GetOrderRequestSteps.cs
```diff
@@ -85,6 +85,12 @@ public void ThenTheGetOrderResponseContainsYearlyValue(decimal amount)
             getOrderResponse.AssertOrderItemCost(amount);
         }
 
+        [Then(@"the get order response contains recipient with (.*) a yearly value of (.*)")]
+        public void ThenTheGetOrderResponseContainsRecipientWithAYearlyValueOf(string odsCode, decimal amount)
+        {
```
bjss-chris, line 90: As per the comments above, I think it might be better if we take a catalogue item ID parameter here.

bjss-chris (changes requested):
Changes look good and the new scenario is good addition. However, I've suggested a change to one of the steps that I think will be an improvement.

Take a look and let me know if you think there might be any problems.

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Order Get Calculations.feature
RahulJha-mastek: Yes that will add filter functionality on `CatalogueItemId` level.
I have rephrased scenario step and also updated steps accordingly.

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Responses/GetOrderResponse.cs
RahulJha-mastek: Scenario steps updated with CatalogueItemId.
If CatalogueItemName is NOT NULL then based on that I have also updated metho `AssertOrderItemCost` to get the actual total cost

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Steps/GetOrderRequestSteps.cs
RahulJha-mastek: Added `catalogueItemID` parameter in step method.

bjss-chris (approved):

PR #380, opened 2021-04-12
Update Order cost calculations

RahulJha-mastek:
Story AB#131476
Task [AB#13692](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13692)


RahulJha-mastek:
Changes not needs to be pushed

PR #385, opened 2021-04-13
ALL Catalogue Solutions deleted

kiran-mastek:
Set Funding Source incomplete if all items deleted

Story: [AB#9039](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9039)
Task: [AB#13738](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13738)

### DIFF
--- a/src/NHSD.BuyingCatalogue.Ordering.Api/Extensions/OrderExtensions.cs
+++ b/src/NHSD.BuyingCatalogue.Ordering.Api/Extensions/OrderExtensions.cs
@@ -36,7 +36,7 @@ namespace NHSD.BuyingCatalogue.Ordering.Api.Extensions
 
         public static bool IsFundingSourceComplete(this Order order)
         {
-            return order?.FundingSourceOnlyGms is not null && (order.HasSolution() || order.HasAdditionalService());
+            return order?.FundingSourceOnlyGms is not null && (order.HasSolution() || order.HasAssociatedService());
         }
 
         public static bool IsAssociatedServicesSectionComplete(this Order order)
--- a/src/NHSD.BuyingCatalogue.Ordering.Api/Extensions/OrderExtensions.cs
+++ b/src/NHSD.BuyingCatalogue.Ordering.Api/Extensions/OrderExtensions.cs
@@ -36,7 +36,7 @@ namespace NHSD.BuyingCatalogue.Ordering.Api.Extensions
 
         public static bool IsFundingSourceComplete(this Order order)
         {
-            return order?.FundingSourceOnlyGms is not null;
+            return order?.FundingSourceOnlyGms is not null && (order.HasSolution() || order.HasAdditionalService());
         }
 
         public static bool IsAssociatedServicesSectionComplete(this Order order)
--- a/src/NHSD.BuyingCatalogue.Ordering.Domain/Order.cs
+++ b/src/NHSD.BuyingCatalogue.Ordering.Domain/Order.cs
@@ -206,9 +206,9 @@ namespace NHSD.BuyingCatalogue.Ordering.Domain
                 || o.CatalogueItem.ParentCatalogueItemId == catalogueItemId);
         }
 
-        public bool HasAdditionalService()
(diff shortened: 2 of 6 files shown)


src/NHSD.BuyingCatalogue.Ordering.Api/Extensions/OrderExtensions.cs
```diff
@@ -36,7 +36,7 @@ public static bool IsAdditionalServicesSectionComplete(this Order order)
 
         public static bool IsFundingSourceComplete(this Order order)
         {
-            return order?.FundingSourceOnlyGms is not null;
+            return order?.FundingSourceOnlyGms is not null && (order.HasSolution() || order.HasAdditionalService());
```
bjss-chris, line 39: An order can only have an additional service if it has a solution. However, an order can contain only an associated service (and no solutions) so I think we need `HasAssociatedService` instead of `HasAdditionalService`.

src/NHSD.BuyingCatalogue.Ordering.Domain/Order.cs
```diff
@@ -206,6 +206,11 @@ public int DeleteOrderItem(CatalogueItemId catalogueItemId)
                 || o.CatalogueItem.ParentCatalogueItemId == catalogueItemId);
         }
 
+        public bool HasAdditionalService()
+        {
+            return OrderItems.Any(o => o.CatalogueItem.CatalogueItemType.Equals(CatalogueItemType.AdditionalService));
+        }
```
bjss-chris, line 212: I think this should be `HasAssociatedService` (see above).

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Order Summary.feature
```diff
@@ -130,12 +130,12 @@ Scenario: get the order summary after a section has been viewed
...
+        | True                      | False                      | False                      | False                      | NULL                    | complete                  | incomplete                 | incomplete                 | incomplete                 | incomplete              |
+        | False                     | True                       | False                      | False                      | NULL                    | incomplete                | incomplete                 | incomplete                 | incomplete                 | incomplete              |
+        | False                     | False                      | True                       | False                      | NULL                    | incomplete                | incomplete                 | complete                   | incomplete                 | incomplete              |
+        | False                     | False                      | False                      | True                       | False                   | incomplete                | incomplete                 | incomplete                 | complete                   | incomplete              |
```
bjss-chris, line 137: I believe this (penultimate) example was correct previously, i.e. funding source section should be complete when associated services section is viewed and the funding source is set.

bjss-chris (changes requested):

tests/NHSD.BuyingCatalogue.Ordering.Api.IntegrationTests/Features/Order Summary.feature
kiran-mastek: Is that the case even if there's no order item added to the order - all lists are empty in this test:
image

bjss-chris: Ah, apologies – I'd missed that. In that case I believe your change is correct.

src/NHSD.BuyingCatalogue.Ordering.Api/Extensions/OrderExtensions.cs
kiran-mastek: Doh! apologies, confused Additional & Associated services - corrected.

src/NHSD.BuyingCatalogue.Ordering.Domain/Order.cs
kiran-mastek: corrected

bjss-chris (approved):

PR #393, opened 2021-04-20
Order item costperyear fix for OnDemand provision type

RahulJha-mastek:
Story [AB#13476](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13476)
Task [AB#13875](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13875)


bjss-chris (approved):
I think we should look at adding some unit tests for `OrderModel` but let's create a separate task for that.

PR #399, opened 2021-04-21
Move DB functionality to service layer – SectionStatusController

RahulJha-mastek:
Task [AB#13834](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13834)


src/NHSD.BuyingCatalogue.Ordering.Services/SectionStatusService.cs
```diff
@@ -0,0 +1,47 @@
...
+            if (sectionId is null)
+                throw new ArgumentNullException(nameof(sectionId));
+
+            var progressUpdate = sectionId switch
+            {
+                "catalogue-solutions" => order.Progress.CatalogueSolutionsViewed = true,
+                "additional-services" => order.Progress.AdditionalServicesViewed = true,
+                "associated-services" => order.Progress.AssociatedServicesViewed = true,
+                _ => throw new InvalidOperationException(),
+            };
```
bjss-chris, line 42: Although the `switch` expression syntax is nicer, in this case I think we should stick with a `switch` statement as we're not actually returning a value or using the `progressUpdate` variable. We could use a discard but I think it would be clearer just to use the `switch` statement.

tests/NHSD.BuyingCatalogue.Ordering.Services.UnitTests/DefaultDeliveryDateServiceTests.cs
```diff
@@ -82,15 +83,38 @@ public static void Constructor_NullHttpContextAccessor_ThrowsArgumentNullExcepti
...
+            [Frozen] CallOffId callOffId,
+            [Frozen] CatalogueItemId catalogueItemId,
+            DateTime defaultDeliveryDate,
+            Order order,
+            DefaultDeliveryDateService service)
+        {
+            context.Order.Add(order);
+            await context.SaveChangesAsync();
+
+            DeliveryDateResult result = await service.SetDefaultDeliveryDate(callOffId, catalogueItemId, defaultDeliveryDate.AddDays(-1));
```
bjss-chris, line 108: `result` isn't used so we can remove the assignment.

bjss-chris (changes requested):
Looks good – just a couple of small changes requested.

src/NHSD.BuyingCatalogue.Ordering.Services/SectionStatusService.cs
RahulJha-mastek: Used Switch statement.

tests/NHSD.BuyingCatalogue.Ordering.Services.UnitTests/DefaultDeliveryDateServiceTests.cs
RahulJha-mastek: Removed `result` assignment

bjss-chris (approved):

PR #401, opened 2021-04-22
Catalogue Solution Variable OnDemand Estimation Period

kiran-mastek:
Remove Estimation Period from CSV file

Story: AB#14030
Task: AB#14031


bjss-chris (commented):
Changes look good.

We have some integration test scenarios that cover the CSV file in the Order Update Status feature but I don't think this particular scenario is currently covered. Is it possible to add a scenario or update one of the existing ones to cover this?

bjss-chris (approved):

PR #407, opened 2021-04-26
Update User Authorization to include RelatedOrganisationId's in check

leerogers-bjss:
Story: [AB#12958](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/12958)
Task: [AB#13895](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13895)

### DIFF
--- a/src/NHSD.BuyingCatalogue.Ordering.Api/Authorization/OrganisationAuthorizationFilter.cs
+++ b/src/NHSD.BuyingCatalogue.Ordering.Api/Authorization/OrganisationAuthorizationFilter.cs
@@ -60,7 +60,10 @@ namespace NHSD.BuyingCatalogue.Ordering.Api.Authorization
             if (result is not null)
                 return (false, result);
 
-            var userAuthorisedOrganisations = user.FindAll(UserClaimTypes.RelatedOrganisationId).Select(c => c.Value).Append(user.FindFirstValue(UserClaimTypes.PrimaryOrganisationId)).ToList();
+            var userAuthorisedOrganisations = user.FindAll(UserClaimTypes.RelatedOrganisationId)
+                                                  .Select(c => c.Value)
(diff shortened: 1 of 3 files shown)


kiran-mastek (changes requested):
Business logic (types of claims etc) should be accessible via the service layer without calling classes needing to be aware of implementation details.

Depending on time scales, I'd highly recommend pulling the logic for needing access to Claims Principal into the `IIdentityService` or a new service & injecting it into the filter - no need for private/static methods etc which allows easier unit testing, besides the ease of reuse & update.

leerogers-bjss:
> Business logic (types of claims etc) should be accessible via the service layer without calling classes needing to be aware of implementation details.
> 
> Depending on time scales, I'd highly recommend pulling the logic for needing access to Claims Principal into the `IIdentityService` or a new service & injecting it into the filter - no need for private/static methods etc which allows easier unit testing, besides the ease of reuse & update.

i would be hesitant in making any major changes at the moment, since all this logic (including JWT tokens) will be removed with the new version of the Application. This is more of a stop gap change until that time

bjss-chris (approved):

bjss-chris:
> Business logic (types of claims etc) should be accessible via the service layer without calling classes needing to be aware of implementation details.
> 
> Depending on time scales, I'd highly recommend pulling the logic for needing access to Claims Principal into the `IIdentityService` or a new service & injecting it into the filter - no need for private/static methods etc which allows easier unit testing, besides the ease of reuse & update.

As Lee's suggested, I'm not sure we'd gain much value from this right now. However, I think it's something we might want to look at when we start work on the order approval process.

kiran-mastek (approved):

PR #435, opened 2021-05-14
Remove ServiceRecipientsViewed from OrderProgress table

RahulJha-mastek:
Task [AB#14677](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/14677)


src/NHSD.BuyingCatalogue.Ordering.Domain/Order.cs
```diff
@@ -181,7 +164,6 @@ public bool CanComplete()
                 && associatedServicesCount > 0;
 
             var recipientsAndAssociatedServices = serviceRecipientsCount == 0
-                && Progress.ServiceRecipientsViewed
                 && associatedServicesCount > 0;
```
bjss-chris, line 167: We should be able to remove `recipientsAndAssociatedServices` completely now – an associated services-only order should be covered by the `noSolutionsAndAssociatedServices` condition now.

We should also be able to remove `serviceRecipientsCount` too as model validation ensures that an order item must have at least one recipient. Obviously, we'll need to remove `serviceRecipientsCount > 0` from the `noSolutionsAndAssociatedServices` condition.

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Extensions/OrderExtensionsTests.cs
```diff
@@ -213,16 +189,15 @@ public static void IsSectionStatusCompleteComplete_NullOrder_ReturnsFalse()
...
-        [CommonInlineAutoData(true, true, false, false, false, false, false)]
-        [CommonInlineAutoData(true, true, false, true, true, false, false)]
-        [CommonInlineAutoData(true, false, true, false, false, true, false)]
+        [CommonInlineAutoData(null, false, false, false, false, false)]
+        [CommonInlineAutoData(true, true, true, true, false, true)]
+        [CommonInlineAutoData(true, true, true, true, true, true)]
+        [CommonInlineAutoData(true, true, true, false, true, true)]
+        [CommonInlineAutoData(true, false, false, false, false, false)]
+        [CommonInlineAutoData(true, false, true, true, false, false)]
+        [CommonInlineAutoData(true, true, false, false, true, true)]
```
bjss-chris, line 198: I think the expected result should still be `false` here, as (if I remember correctly) a buyer must have viewed the solutions page before they can complete an order. Worth double-checking.

tests/NHSD.BuyingCatalogue.Ordering.Api.UnitTests/Models/Summary/OrderSummaryModelTests.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
```
bjss-chris, line 4: We shouldn't need to add `System.Linq` and I think we can remove `System.Collections.Generic` too.

src/NHSD.BuyingCatalogue.Ordering.Api/Models/ServiceRecipientModel.cs
```diff
@@ -10,12 +10,6 @@ public ServiceRecipientModel()
         {
         }
```
bjss-chris, line 11: We can remove the empty constructor and `using NHSD.BuyingCatalogue.Ordering.Domain;` now.

bjss-chris (changes requested):
Looks good on the whole. However, I think there is a little extra code that we can remove and I think one of the unit test cases may now be incorrect.

bjss-chris (approved):
👍

PR #446, opened 2021-05-25
Enable Integration test for updating LastUpdated information

RahulJha-mastek:
Task [AB#14905](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/14905)

Enable Integration test for updating LastUpdated information in Order, when OrderItem is updated.


bjss-chris (commented):
Rather than introducing the identity service into `CreateOrderItemService` can we just update `lastUpdated` in `Order.AddOrUpdateOrderItem`?

This should cause EF Core tracking to mark the `Order` as changed, which will invoke `SetLastUpdatedBy` when `SaveChangesAsync` is invoked.

It might be worth adding a comment, assuming that works, to explain why we're doing it. For example:

```c#
// Included to force EF Core to track order as changed so audit information is updated
lastUpdated = DateTime.UtcNow;
```

